PR #43, opened 2020-10-09
Add model validation

simongraband:
Endpoints added:
- validate (Validates once)
- validation/subscribe (Enables live validation)
- validation/constraints (Returns a list of Constraints that can be used for Input Validation)

Signed-off-by: Simon Graband <simon.graband@tum.de>


ndoschek (changes requested):
Thank you very much for your PR Simon! The validation mechanism is a great contribution to the model server.
The main issue I noticed however, is the subscription mechanism which is not fully complete yet, e.g. it does not send updates if models are updated via commands (please see my comments inline).
I have noticed some other issues at the moment and please be sure to rebase the pull-request and resolve the conflicts.

Also, I have another request regarding the [`emfcloud-modelserver-theia` repository](https://github.com/eclipse-emfcloud/emfcloud-modelserver-theia/tree/master/modelserver-theia): Please extend the `ModelServerClient` and the `ModelServerSubscriptionClient` accordingly and add small examples to the test menus in the [`dev-example`](https://github.com/eclipse-emfcloud/emfcloud-modelserver-theia/tree/master/example/dev-example).

ndoschek (changes requested):
Thank you Simon for your changes! I added some new comments inline, which target mainly some implementation details. Please also make sure again to rebase to the current master branch.

In general it would be great to detach the EMF validation functionality into a separate `ModelValidator` service which is then configurable with DI and is then injected by the `ModelController` (which is basically a redirector and does not hold a lot of code logic itself).
Basically you would need to extract a generic interface that offers the basic methods and use your `EMFModelValidator` (which then implemenents said interface) as default binding for the `ModelValidator`. As reference you can look at the concept of `ModelResourceManager` and its default implementation `DefaultModelResourceManager`.

As we removed the separate websocket connection for validation updates, while testing I thought of a configuration possibility for the validation subscription updates:
- We could add a 'liveValidation' flag in form of an optional url parameter (e.g. something like `&liveValidation=true` which defaults to `false` if not set) and handle it in a similar way like the `format` parameter. That means on broadcasting the `validationResult` we would notify only connections with the flag set explicitly to `true`. What do you think about that?

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/ValidationMapperModule.java
```diff
@@ -0,0 +1,159 @@
...
+
+      @Override
+      public void serialize(final BasicDiagnostic value, final JsonGenerator gen, final SerializerProvider serializers)
+         throws IOException {
+         gen.writeStartObject();
+         Optional<EObject> eObject = value.getData().stream()
+            .filter(EObject.class::isInstance)
+            .map(EObject.class::cast)
+            .findFirst();
+         gen.writeObjectField("id", res.getURIFragment((eObject.get())));
```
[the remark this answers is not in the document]
simongraband: Yes, i agree. I added them in the ValidationMapperModule as public constants. Is this fine or would you propose another location?

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/SessionController.java
```diff
@@ -43,6 +47,8 @@
 
    private final Map<String, Set<WsContext>> modelUrisToClients = Maps.newConcurrentMap();
 
+   private final Map<String, BasicDiagnostic> modelUriToLastSendDiagnostic = Maps.newConcurrentMap();
+
```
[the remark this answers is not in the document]
simongraband: This is used to avoid sending a Client an update if the previous send diagnostic is equal to the new one. I can also remove it if this is not wanted.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/ModelController.java
```diff
@@ -159,6 +170,7 @@ public void update(final Context ctx, final String modeluri) {
                } catch (EncodingException e) {
                   handleEncodingError(ctx, e);
                }
+               this.modelRepository.validate(modeluri);
```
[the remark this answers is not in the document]
simongraband: undo and redo make sense for sure. I will add them. 
However, for save and when the model is manually validated i am not sure. You could argue, that when the model is saved, the model is not changing therefore the validation result will not be different than before. 
So broadcasting the validation result while saving/manually validating will act like a refresh, that may be able to solve some problems that happened before, but if nothing went wrong it will resend the same results to all the clients. 
Which way would you prefer?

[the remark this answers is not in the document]
ndoschek: Yes you're totally right. I just took a deeper look at the validation behavior in the tree model editor in Eclipse for comparison:
- If live validation is activated, there is no need to send an additional validation update on saving.
- If live validation is deactivated, the client/user is responsible on its own to trigger validation anyway. In Eclipse it is also not triggered automatically on `save`, so therefore it's fine that way I would say :)

simongraband:
I updated the PR according to your feedback.

The validation mechanism and the `ObjectMapper` are now configurable via DI.

Furthermore, I introduced a liveValidation parameter for the subscribe endpoint, so that each websocket can turn the validation updates on or off. The parameter defaults to `false`, where it will not receive updates on the validation result. A re-validation is triggered by a change on the model or an undo/redo operation. If the parameter is set to `true` the websocket will receive the updated validation result, if it is different form the previously sent one.

ndoschek (changes requested):
Thanks again for your changes Simon! The validation framework works great now and I just have a few minor comments, but other than we are on a good way to merge it.

Regarding my former request, it would be great to make the new API methods available at @eclipse-emfcloud/modelserver-theia (along with a very basic example maybe) such that everyone can see the validation mechanism in action which already looks great in code.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/DefaultModelValidator.java
```diff
@@ -0,0 +1,138 @@
...
+   @Inject
+   public DefaultModelValidator(final ModelRepository modelRepository, final FacetConfig facetConfig) {
+      this.modelRepository = modelRepository;
+      this.facetConfig = facetConfig;
+   }
+
+   @Override
+   public JsonNode validate(final String modeluri) {
+      Optional<EObject> model = this.modelRepository.getModel(modeluri);
+      if (model.isPresent()) {
```
ndoschek, line 52: Do we need this outer check here? Isn't it enough to check if the resource exists?

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/EMFFacetConstraints.java
```diff
@@ -0,0 +1,171 @@
...
+   public void setMinInclusive(final String minInclusive) { this.minInclusive = minInclusive; }
+
+   public String getMaxInclusive() { return maxInclusive; }
+
+   public void setMaxInclusive(final String maxInclusive) { this.maxInclusive = maxInclusive; }
+
+   public EMFFacetConstraints(final Map<String, Object> facetMap) {
+      this.whiteSpace = (Integer) facetMap.getOrDefault(WHITESPACE, null);
+      this.enumeration = (List<String>) facetMap.getOrDefault(ENUMERATION, null);
+      this.pattern = (List<String>) facetMap.getOrDefault(PATTERN, null);
```
ndoschek, line 107:
```suggestion
   @SuppressWarnings("unchecked")
   public EMFFacetConstraints(final Map<String, Object> facetMap) {
      this.whiteSpace = (Integer) facetMap.getOrDefault(WHITESPACE, null);
      this.enumeration = (List<String>) facetMap.getOrDefault(ENUMERATION, null);
      this.pattern = (List<String>) facetMap.getOrDefault(PATTERN, null);
```

Please void warning here

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/EMFFacetConstraints.java
```diff
@@ -0,0 +1,171 @@
...
+   public boolean equals(final Object obj) {
+      if (obj instanceof Integer) {
+         return obj == Integer.valueOf(0);
+      }
+      return true;
+   }
+
+   @Override
+   public int hashCode() {
+      // TODO Auto-generated method stub
```
ndoschek, line 136:
```suggestion
```
Can be removed

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/EMFFacetConstraints.java
```diff
@@ -0,0 +1,171 @@
...
+   public boolean equals(final Object obj) {
+      if (obj instanceof Integer) {
+         return obj == Integer.valueOf(-1);
+      }
+      return true;
+   }
+
+   @Override
+   public int hashCode() {
+      // TODO Auto-generated method stub
```
ndoschek, line 152:
```suggestion
```
Can be removed

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/EMFFacetConstraints.java
```diff
@@ -0,0 +1,171 @@
...
+      return super.hashCode();
+   }
+}
+
+class StringListFilter {
+   @Override
+   public boolean equals(final Object obj) {
+      if (obj instanceof List) {
+         return ((List) obj).isEmpty();
```
ndoschek, line 161:
```suggestion
         return ((List<?>) obj).isEmpty();
```
Please avoid warning here

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/EMFFacetConstraints.java
```diff
@@ -0,0 +1,171 @@
...
+      }
+      return true;
+   }
+
+   @Override
+   public int hashCode() {
+      // TODO Auto-generated method stub
```
ndoschek, line 168:
```suggestion
```
Can be removed

ndoschek (commented):
Thanks again for your changes Simon! Looks great to me, I just found some very minor nitpicks.

Also I have one question: I cannot remember anymore if the id of the `validationResult` was always ` "id": "/-1"`. Was there a change, and if yes was this intended?

simongraband:
Hey Nina,
the `"id": "/-1"` is not intended. I tested it again and it seems like this has something to do with the `Resource` not being loaded correctly, since the '`"id": "/-1"` only happens, when validating a model after another model has already been validated. Going back to the original model works as expected. I will look into it.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/DefaultModelValidator.java
```diff
@@ -0,0 +1,136 @@
...
+   public DefaultModelValidator(final ModelRepository modelRepository, final FacetConfig facetConfig) {
+      this.modelRepository = modelRepository;
+      this.facetConfig = facetConfig;
+   }
+
+   @Override
+   public JsonNode validate(final String modeluri, final ObjectMapper mapper) {
+      Optional<EObject> model = this.modelRepository.getModel(modeluri);
+      Optional<Resource> res = this.modelRepository.loadResource(modeluri);
+      if (res.isPresent()) {
```
ndoschek, line 53: Ah sorry, my bad! I oversaw that you use the `model` variable below... Could you just combine the checks in this if clause so we are on the safe side!
```suggestion
      if (model.isPresent() && res.isPresent()) {
```

ndoschek (changes requested):

ndoschek (approved):
Thank you again Simon for addressing my numerous requests and especially for contributing the validation mechanism, it's a great extension to the model server! :tada:

PR #46, opened 2020-10-14
Fix #45: Ensure UI Schema directory is skipped when scanning workspace

### LINKED ISSUE #45: GetAll request is failing
Requesting all available models from the model server via "GetAll" (http://localhost:8081/api/v1/models) fails with an internal server error:
```
WARN io.javalin.Javalin  - Uncaught exception
org.eclipse.emf.common.util.BasicEList$BasicIndexOutOfBoundsException: index=0, size=0 ...
```

martin-fleck-at:
- Make sure the UI schema directory is part of the server configuration
- Skip UI schema directory resources when loading workspace resources
- Guard against NPE by checking resource content in 'getAllModels'

### DIFF
--- a/bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/ModelRepository.java
+++ b/bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/ModelRepository.java
@@ -104,9 +104,7 @@ public class ModelRepository {
    }
 
    protected boolean isSourceDirectory(final File file) {
-      System.out.println("isSource[" + file.getAbsolutePath() + "] " + file.isDirectory() + " - "
-         + (!this.serverConfiguration.isUISchemaFolder(file.getAbsolutePath())));
-      return file.isDirectory() && !this.serverConfiguration.isUISchemaFolder(file.getAbsolutePath());
+      return file.isDirectory() && !this.serverConfiguration.isUiSchemaFolder(file.getAbsolutePath());
    }
 
    protected boolean hasModel(final String modeluri) {
--- a/bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/ModelRepository.java
+++ b/bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/ModelRepository.java
@@ -66,25 +66,34 @@ public class ModelRepository {
       this.domain = new AdapterFactoryEditingDomain(adapterFactory, new BasicCommandStack(), resourceSet);
       this.serverConfiguration = serverConfiguration;
       this.resourceManager = resourceManager;
-      initialize(serverConfiguration.getWorkspaceRootURI().toFileString(), true);
+      initialize();
    }
 
-   public void initialize(final String workspaceRoot, final boolean clearResources) {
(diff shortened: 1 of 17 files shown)


bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/configuration/ServerConfiguration.java
```diff
@@ -85,6 +89,12 @@ public void setUISchemaFolder(final String uiSchemaFolder) {
 
    public void setServerPort(final int serverPort) { this.serverPort = serverPort; }
 
+   @Override
+   public String toString() {
+      return "ServerConfiguration [workspaceRootURI=" + workspaceRootURI + ", uiSchemaFolderURI=" + uiSchemaFolderURI
+         + ", serverPort=" + serverPort + "]";
+   }
+
```
ndoschek, line 97: Why do we need this?

ndoschek (changes requested):
Hi Martin! Thank you very much for your changes regarding the UI schema folder issues. They work really well and now it is more clear that the client can set its own UI schema folder or can use the one from the Temp modelserver workspace. :+1: 

However I think we should decide on one consistent notation for UI/Ui schema. As far as I can tell you used UiSchema throughout your changes, but there are still some occurences where UISchema is used (e.g. in the `ModelServerClient`). It would be great if you could equalize those, I'm fine with either one of the two spellings.
Other than that I found two very minor issues, please see inline.

Thank you again for your great change!

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/ModelRepository.java
```diff
@@ -94,7 +103,13 @@ public void initialize(final String workspaceRoot, final boolean clearResources)
       }
    }
 
-   boolean hasModel(final String modeluri) {
+   protected boolean isSourceDirectory(final File file) {
+      System.out.println("isSource[" + file.getAbsolutePath() + "] " + file.isDirectory() + " - "
+         + (!this.serverConfiguration.isUISchemaFolder(file.getAbsolutePath())));
```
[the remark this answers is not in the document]
martin-fleck-at: Yeah, we do not need that at all, that was just a leftover from my local testing. Sorry for that!

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/configuration/ServerConfiguration.java
martin-fleck-at: It is not strictly necessary but I found it useful during debugging and it also shouldn't do any harm. So if you don't have any strong objections I would argue to keep that :-)

martin-fleck-at:
Thank you very much for your thorough and kind review, Nina! I absolutely agree with you that we should have a consistent use of the term 'UiSchema' so in a follow-up commit I adapted all the usages I could find. This also had an effect on the related change in https://github.com/eclipse-emfcloud/emfcloud-modelserver-theia/pull/18. 

I also added a test case for the newly added 'isUiSchemaFolder' method. If you have some time it would be great if you could take another look. Thank you so much!

ndoschek (approved):
Thank you again Martin for your efforts, looks great to me! :+1:

PR #47, opened 2020-11-03
Add necessary dependencies to modelserver.lib bundle

eneufeld:
When trying to use the org.eclipse.emfcloud.modelserver.lib as a bundle,
the libraries inside the bundle can't resolve necessary dependencies, eg
slf4j.
By adding the dependencies explicitly to the MANIFEST, this issue is
resolved.

Signed-off-by: Eugen Neufeld <eneufeld@eclipsesource.com>

### DIFF
--- a/bundles/org.eclipse.emfcloud.modelserver.lib/.classpath
+++ b/bundles/org.eclipse.emfcloud.modelserver.lib/.classpath
@@ -6,7 +6,6 @@
 		</attributes>
 	</classpathentry>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
-	<classpathentry kind="src" path="src/"/>
 	<classpathentry exported="true" kind="lib" path="lib/annotations-13.0.jar"/>
 	<classpathentry exported="true" kind="lib" path="lib/kotlin-stdlib-1.3.40.jar"/>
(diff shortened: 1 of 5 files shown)


tortmayr:
@eneufeld Looks CI build is failing because you added jetty as a plugin dependency but did not add it to the target configuration.

(https://github.com/eclipse-glsp/glsp-server/blob/d8c4f6989bf25e64ffd80dd36c61b3dde6dbe8ef/targetplatforms/r2020-06.tpd#L22)

CamilleLetavernier (commented):
I would add org.slf4j.api as a required-bundle (rather than simply importing org.slf4j), to avoid this runtime warning:

```
-------------------------------------------------------------------
Missing dependency 'Slf4j simple'. Add the dependency.

pom.xml:
<dependency>
    <groupId>org.slf4j</groupId>
    <artifactId>slf4j-simple</artifactId>
    <version>1.7.26</version>
</dependency>

build.gradle:
compile "org.slf4j:slf4j-simple:1.7.26"
-------------------------------------------------------------------
Visit https://javalin.io/documentation#logging if you need more help
```

There's also an (unrelated) issue in the .classpath, which references a src/ folder that doesn't exist (Same thing for build.properties, which contains both source.. and output.. entries, even though they don't exist)

Issue #40, opened 2020-06-10
Typeschema does not return all information


cbos:
We can request a typeschema via a request to this url: api/v1/typeschema?modeluri=...

But the returned schema, only returns information about the direct class, but not information of the super class.
We have a test model where we have a common ecore model with interfaces, and we have 2 derived ecore models. The specific field attributes or relations defined in the derived model are returned, but the attibutes and relations in the common model are not returned even no references.
Due to that the returned typeschema is incomplete. 

Either it should return a reference to these schemas or it should include it.

ndoschek:
Hi! Sorry for the late reply. You are right, thanks for raising that issue.
Personally I would prefer to add the elements from the common model instead of references in regards to usage, i.e. one request by the consumer is then enough to get the complete type information.
If you have already a solution in mind we would happily accept a contribution for this. Otherwise we will take a look at it in the next weeks.

cbos:
@ndoschek 
We did not make a solution yet. For now we created the typeschemas manually to explore the possibilities of jsonforms. 

If we start working on it we will definitely keep in touch to see how we can contribute it, but we did not plan for that yet.

In case there is a working solution committed by one of you, we can easily update our setup and test it to see if that works based on what we experience right now to give the feedback as soon as possible.

ndoschek:
That sounds good, let's keep in touch on that matter! Thanks!

koegel:
Closing for inactivity, feel free to reopen.

PR #49, opened 2020-11-18
Support Undo/Redo API

### LINKED ISSUE #14: Support Undo/Redo API
- [ ] undo and redo commands on the `edit` endpoint
- [ ] on undo, compute the inverse of the command that was undone and send that to subscribers for incremental update notification
- [ ] on redo, send the redone command to subscribers
- [ ] ensure that the save API tells the `CommandStack` that save was done
- [ ] if undo returns to the last savepoint, notify subscribers about dirty state change (command-stack listener)

### LINKED ISSUE #50: Each resource should be loaded in a separate editing domain
To improve the undo/redo support each resource identified by the model URI should be loaded in a separate editing domain and a separate resource set. All edit operations are handled by the editing domain associated with the respective resource.

This behavior opens up the possibility to have two editing domains/resource sets loading the same model (either directly or through cross references). Therefore, after performing a save operation a notification should be sent out to all listeners of that file that the underlying content has changed.

### LINKED ISSUE #51: Make model loading/creation of editing domain configurable
Currently the model server creates the editing domain, the resource set and loads the model all by itself. However, there might be cases where additional resources should be loaded in the corresponding resource set/editing domain. We therefore should provide an Interface and default implementation for the model loading that is bound in the model server module so that it can be customized. By default each resource should be loaded in a separate editing domain, see implementation provided in issue #50.

ndoschek:
- Each resource should be loaded in a separate editing domain (#50)
  - Introduce ModelResourceLoader, which covers the whole resource handling in favour of ResourceManager and ModelRepository. It is also responsible for the handling of EditingDomains. The DefaultModelResourceLoader creates one ResourceSet and EditingDomain each per Resource.
- Make model loading/creation of editing domain configurable (#51)
  - Introduce DefaultModelServerModule which is the default implementation of the now abstract ModelServerModule. It configures default bindings and will be completed with (#54)
- Undo/Redo API is relying on the two mentioned changes regarding Resource handling. Undo/Redo is covered by the ModelResourceLoader and the default implementation is capable of undo/redo per modeluri and notifies by incremental update by sending the corresponding command to all listeners.
  - In case of the Undo command, it was necessary to invert the UndoCommand, such that the listeners are able to apply the changes without having to implement their own CommandStack concept.
- Adjust UnitTests

Resolves #14, resolves #50, resolves #51

Please also see: https://github.com/eclipse-emfcloud/emfcloud-modelserver-theia/pull/24


eneufeld (commented):
Looks good to me, just have 2 small remarks.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/DefaultModelResourceLoader.java
```diff
@@ -0,0 +1,336 @@
...
+         editingDomain.redo();
+         return redoCommand;
+      }
+      return null;
+   }
+
+   @Override
+   public boolean save(final String modeluri) {
+      ResourceSet resourceSet = getResourceSet(modeluri);
+      boolean result = resourceSet.getResources().stream().allMatch(this::saveResource);
```
martin-fleck-at, line 314: This actually ignores the 'modelUri', see also [1]. I suggest to make the model uri optional and only save the single resource if it is given and all resources otherwise. Or do you think that saving only a single resource may cause problems down the road? Because currently we get the editing domain for the resource and call 'saveIsDone' on it.

[1] https://spectrum.chat/emfcloud/modelserver/save-api-ignores-parameter-modeluri~a9b386f6-80d2-4fbf-a8f9-5fbc354b6a6e

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/ModelResourceLoader.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.eclipse.emf.common.command.Command;
+import org.eclipse.emf.common.util.URI;
+import org.eclipse.emf.ecore.EObject;
+import org.eclipse.emf.ecore.resource.Resource;
+import org.eclipse.emf.ecore.resource.ResourceSet;
+import org.eclipse.emfcloud.modelserver.command.CCommand;
+import org.eclipse.emfcloud.modelserver.common.codecs.DecodingException;
+
+public interface ModelResourceLoader {
```
martin-fleck-at, line 26: Just a thought: Not sure if I would go with 'Loader' here since this class also provides updates, removal, commands, saving etc. So maybe use 'Manager', 'Handler' or something similar instead?

martin-fleck-at (changes requested):
Everything seems to work well, thank you very much for that great work! I just have a few minor comments and a request: I tested this with the SuperBrewer3000.coffee file and it seemed to work well. However, there is no way to easily test the separation of editing domains (i.e., one per resource) since we only provide edit commands for a single file, see my comment in https://github.com/eclipse-emfcloud/emfcloud-modelserver-theia/pull/24. I would love to see in action what already looks great in code.

bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/client/ModelServerClient.java
```diff
@@ -659,4 +659,28 @@ public boolean close(final EditingContext editingContext) {
...
       return false;
    }
+
+   @Override
+   public CompletableFuture<Response<Boolean>> undo() {
+      final Request request = new Request.Builder()
+         .url(createHttpUrlBuilder(makeUrl(UNDO)).build())
+         .build();
+
+      return makeCall(request)
```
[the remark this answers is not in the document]
ndoschek: Thats a good idea, I created several helper methods and introduced additional interfaces for the mentioned constants and also for the query parameters like "modeluri" or "format".
(Remark: I did not do this for the `JsonSchemaConverter` yet, as I will work on an open issue (#40) soon and will work your suggestion there in then as well.)

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/DefaultModelResourceLoader.java
```diff
@@ -0,0 +1,336 @@
...
+
+   private URI createURI(final String modeluri) {
+      return modeluri.startsWith("file:")
+         ? URI.createURI(modeluri, true)
+         : URI.createFileURI(modeluri);
+   }
+
+   @Override
+   @SuppressWarnings("checkstyle:IllegalCatch")
+   public Optional<Resource> loadResource(final String modeluri) {
```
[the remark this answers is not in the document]
ndoschek: I would also stay with the `String` parameter here. As Martin mentions, the other methods accept the `modelUri` also as `String` parameter. This is due to the fact that the correct resolving of the URI is done in this class anyway.

bundles/org.eclipse.emfcloud.modelserver.common/src/org/eclipse/emfcloud/modelserver/common/ModelServerPathsV1.java
```diff
@@ -19,7 +19,9 @@
 
    String SUBSCRIPTION = "subscribe"; // accepts query parameter "modeluri"
    String EDIT = "edit"; // accepts query parameter "modeluri"
-   String SAVE = "save"; // accepts query parameter "modeluri"
+   String SAVE = "save";
```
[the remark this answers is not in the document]
ndoschek: As we have discussed, now we have two paths for saving, `SAVE` which accepts the `modelUri` parameter and `SAVE_ALL` which does not.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/DefaultModelResourceLoader.java
```diff
@@ -0,0 +1,336 @@
...
+import org.eclipse.emfcloud.modelserver.common.codecs.EMFJsonConverter;
+import org.eclipse.emfcloud.modelserver.edit.CommandCodec;
+import org.eclipse.emfcloud.modelserver.emf.configuration.EPackageConfiguration;
+import org.eclipse.emfcloud.modelserver.emf.configuration.ServerConfiguration;
+import org.emfjson.jackson.resource.JsonResourceFactory;
+
+import com.google.common.collect.Maps;
+import com.google.inject.Inject;
+
+public class DefaultModelResourceLoader implements ModelResourceLoader {
```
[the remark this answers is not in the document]
ndoschek: Thanks for noticing!

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/DefaultModelResourceLoader.java
```diff
@@ -0,0 +1,336 @@
...
+   public void updateResource(final String modeluri, final CCommand command) throws DecodingException {
+      Command decoded = commandCodec.decode(getEditingDomain(getResourceSet(modeluri)), command);
+      getEditingDomain(getResourceSet(modeluri)).execute(decoded);
+   }
+
+   @Override
+   public Command undo(final String modeluri) {
+      ModelServerEditingDomain editingDomain = getEditingDomain(getResourceSet(modeluri));
+      Command undoCommand = editingDomain.getCommandStack().getUndoCommand();
+      if (editingDomain.canUndo()) {
```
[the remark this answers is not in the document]
ndoschek: You're totally right, I moved this logic now to the `ModelServerEditingDomain`. :+1:

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/DefaultModelResourceLoader.java
ndoschek: Yes that's true, as we discussed, I now introduced a `saveAll` which saves all `Resource`s of the available `ResourceSet`s. The `save(modelUri)` is fixed now and only saves the dedicated `Resource` behind the `modelUri`.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/ModelResourceLoader.java
ndoschek: I renamed it to `ModelResourceManager`

README.md
```diff
@@ -151,6 +150,10 @@ public interface ModelServerClientApiV1<A> {
    EditingContext edit();
 
    boolean close(EditingContext editingContext);
+
+   CompletableFuture<Response<Boolean>> undo();
+
+   CompletableFuture<Response<Boolean>> redo();
```
martin-fleck-at, line 156: Could you please also add the `saveAll` method to the `save` method?

README.md
```diff
@@ -64,19 +67,14 @@ The following table shows the current HTTP endpoints:
 | |Save|__GET__|`/save`|query parameter: `?modeluri=...`
 | |Execute commands|__PATCH__|`/edit`|query parameter: `?modeluri=...`
 | |Get all available model URIs in the workspace|__GET__|`/modeluris`| -
-| |Get model element by id|__GET__|`/modelelement`|query parameter: `[?modeluri=...&elementid=...[&format=...]]`
-| |Get model element by name <br> (Returns the first element that matches the given `elementname`)|__GET__|`/modelelement`|query parameter: `[?modeluri=...&elementname=...[&format=...]]`
+| |Get model element by id|__GET__|`/modelelement`|query parameter: `?modeluri=...&elementid=...[&format=...]`
```
martin-fleck-at, line 70: Could you please also add the `Save All` endpoint here?

martin-fleck-at (changes requested):
Thank you very, very much Nina! From what I can see everything works very well. I could not find any issues and the additional cleanup is a big bonus! Great work!

I only noticed that the new 'Save All' was missing from the README. If you could also add that I believe we can merge this change!

ndoschek:
Thank you for your review Martin! I updated the README accordingly.

martin-fleck-at (approved):
Great, thank you for the quick update! I'll merge it as soon as the build goes through!

PR #53, opened 2020-11-26
Align Java client API with protocol and typescript client

martin-fleck-at:
- Add missing method for listing model uris: getModelUris
- Ensure getAll returns a list of models and supports custom format
- Introduce dedicated model class with modelUri and content properties
- Add client test cases for new and changed methods
- Fix setup of modelserver.lib


bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/client/ModelServerClientApiV1.java
```diff
@@ -22,7 +23,11 @@
 
    CompletableFuture<Response<A>> get(String modelUri, String format);
 
-   CompletableFuture<Response<List<String>>> getAll();
+   CompletableFuture<Response<List<Model<String>>>> getAll();
+
+   CompletableFuture<Response<List<Model<EObject>>>> getAll(String format);
```
ndoschek, line 28: We should keep the generic return type `<A>` in this case, as the current `ModelServerClient` implements the `ModelServerClientApi<EObject>` anyway.

README.md
```diff
@@ -100,7 +100,11 @@ public interface ModelServerClientApiV1<A> {
 
    CompletableFuture<Response<A>> get(String modelUri, String format);
 
-   CompletableFuture<Response<List<String>>> getAll();
+   CompletableFuture<Response<List<Model<String>>>> getAll();
+
+   CompletableFuture<Response<List<Model<EObject>>>> getAll(String format);
```
ndoschek, line 105: Keep generic type `<A>` here as well.

ndoschek (changes requested):
Thank you Martin for this change! :+1: Code looks good to me, I only noticed one minor API return type mismatch, would be great if you could have a look at it, thanks!

martin-fleck-at:
@ndoschek Thank you for your review, Nina! You are absolutely right, I pushed a follow-up commit that should fix it.

ndoschek (approved):

PR #55, opened 2020-12-04
Improve WebSocket connections regarding timeout and keepAlive messages

ndoschek:
- Allow clients to set idle timeout on subscription (default is 300000ms)
- Handle ws messages (type keepAlive) to enable clients to prolong connection independently
- Extend SessionControllerTest
 - Remove TestOnly predicate in SessionController and mock fields directly
- Remove format argument in subscribe(..) in ModelServerClientAPI as it was never used or implemented (may open an issue for that though)

Please also see https://github.com/eclipse-emfcloud/emfcloud-modelserver-theia/pull/23

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,11 @@ public interface ModelServerClientApiV1<A> {
 
    CompletableFuture<Response<Boolean>> edit(String modelUri, CCommand command, String format);
 
-   void subscribe(String modelUri, SubscriptionListener subscriptionListener, String format);
+   void subscribe(String modelUri, SubscriptionListener subscriptionListener);
+
+   void subscribe(String modelUri, SubscriptionListener subscriptionListener, long timeout);
+
+   boolean send(String modelUri, String message);
 
    boolean unsubscribe(String modelUri);
 
--- a/bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/client/ModelServerClient.java
+++ b/bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/client/ModelServerClient.java
@@ -485,7 +485,7 @@ public class ModelServerClient implements ModelServerClientApi<EObject>, ModelSe
                   .toString()))
          .build();
 
-      subscribe(modelUri, subscriptionListener, request);
+      doSubscribe(modelUri, subscriptionListener, request);
    }
 
    @Override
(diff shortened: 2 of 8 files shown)


bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/client/ModelServerClient.java
```diff
@@ -475,18 +475,37 @@ private boolean isSupportedFormat(final String format) {
...
-                  .addQueryParameter("format", checkedFormat)
+                  .addQueryParameter("timeout", String.valueOf(timeout))
                   .build()
                   .toString()))
          .build();
 
+      subscribe(modelUri, subscriptionListener, request);
+   }
+
+   private void subscribe(final String modelUri, final SubscriptionListener subscriptionListener,
```
CamilleLetavernier, line 507: You should avoid using the same method name for public and private methods; it can confuse the compiler (Especially when the private method uses non-API types, such as `Request` here). You can rename this to e.g. doSubscribe()

Otherwise, client-applications get errors such as:

`The type okhttp3.Request cannot be resolved. It is indirectly referenced from required .class files`

README.md
```diff
@@ -140,7 +140,11 @@ public interface ModelServerClientApiV1<A> {
 
    CompletableFuture<Response<Boolean>> edit(String modelUri, CCommand command, String format);
 
-   void subscribe(String modelUri, SubscriptionListener subscriptionListener, String format);
+   void subscribe(String modelUri, SubscriptionListener subscriptionListener);
```
CamilleLetavernier, line 143: Was it intended to remove the `String format` parameter from the method?

bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/client/ModelServerClient.java
ndoschek: Thank you for your feedback Camille, I addressed this issue now!

README.md
ndoschek: Yes, as mentioned in the description of this PR, I removed that argument as it was never offered by the REST API, and only introduced here in the `ModelServerClientApiV1`. But we could open an issue for implementing the feature of enabling subscription on a certain format though.

CamilleLetavernier: Oops, I didn't read that :) I simply noticed that Ecore GLSP had a compile error with this change.
All good then :+1:

martin-fleck-at (approved):
Change looks good to me now, thank you very much, Nina!

PR #56, opened 2021-01-07
Fix subscription connections and undo/redo calls in ClientAPI

ndoschek:
- Update dependencies

- Fix format parameter, such that subscriptions work correctly again for a requested format (was falsely removed with 96085a0)
- Add missing parameter modelUri to undo/redo calls in ClientAPI
- Add basic client subscription example
- Add ModelServerClientTests for undo, redo and save calls
- Remove warning in SessionControllerTest
- Update README

Please also see https://github.com/eclipse-emfcloud/emfcloud-modelserver-theia/pull/27 which adds an example usage of subscribing to a model in a certain format.


tests/org.eclipse.emfcloud.modelserver.emf.tests/src/org/eclipse/emfcloud/modelserver/emf/common/SessionControllerTest.java
```diff
@@ -48,7 +48,6 @@
 import io.javalin.websocket.WsContext;
 import io.javalin.websocket.WsMessageContext;
 
-@SuppressWarnings("restriction")
 @RunWith(MockitoJUnitRunner.class)
```
martin-fleck-at, line 51:
```suggestion
@SuppressWarnings("restriction")
@RunWith(MockitoJUnitRunner.class)
```

martin-fleck-at (approved):
Change looks good to me in general and everything still seems to work as expected, thank you Nina!

I am not 100% certain what the purpose of those example clients is. Isn't that what is usually covered in test cases or is it more for documentation purposes? Maybe we can clarify that a bit to ensure that they are in the right place but in my opinion that should not prevent us from merging this change.

Same goes for the suppression of the 'restriction' warning. I do get those warnings so I am not sure why you deleted but I suggest to keep the suppression for now. Still that does not prevent merging in my opinion.

tests/org.eclipse.emfcloud.modelserver.emf.tests/src/org/eclipse/emfcloud/modelserver/emf/common/SessionControllerTest.java
```diff
@@ -48,6 +48,7 @@
 import io.javalin.websocket.WsContext;
 import io.javalin.websocket.WsMessageContext;
 
+@SuppressWarnings("restriction")
```
martin-fleck-at: Remove suppress warning, committed by mistake.

martin-fleck-at:
```suggestion
```

martin-fleck-at (approved):

Issue #63, opened 2021-01-13
Mocking WsContext.pathParam sometimes results in WrongTypeOfReturnValue exception


martin-fleck-at:
Since updating to the latest version of Kotlin in https://github.com/eclipse-emfcloud/emfcloud-modelserver/pull/56 the JUnit tests seem to fail **sometimes** with the following error:
```
org.mockito.exceptions.misusing.WrongTypeOfReturnValue: 
String cannot be returned by pathParam()
pathParam() should return Validator
***
If you're unsure why you're getting above error read on.
Due to the nature of the syntax above problem might occur because:
1. This exception *might* occur in wrongly written multi-threaded tests.
   Please refer to Mockito FAQ on limitations of concurrency testing.
2. A spy is stubbed using when(spy.foo()).then() syntax. It is safer to stub spies - 
   - with doReturn|Throw() family of methods. More in javadocs for Mockito.spy() method.

	at org.eclipse.emfcloud.modelserver.emf.common.SessionControllerTest.initializeWsMessageContext(SessionControllerTest.java:190)
	at org.eclipse.emfcloud.modelserver.emf.common.SessionControllerTest.testSubscribeAndSendKeepAlive(SessionControllerTest.java:159)
```

After investigating that for a bit, it seems that there was a change that marked the method `pathParam` with `@JvmSynthetic` to might have fixed that but did not in our case: https://github.com/tipsy/javalin/commit/be8910b3e867e1662663fbedb4e914c4262aff64.

There are already some open issues related to this:
- https://github.com/tipsy/javalin/issues/979
- https://github.com/tipsy/javalin/issues/1098
- https://github.com/mockito/mockito/issues/1943

The problem should be investigated and fixed, either by re-writing the mocking code, downgrading to a Javalin version that works (previously we used 3.1.0) of finding another way to get the test cases to run properly.

martin-fleck-at:
For the moment, we downgraded to Javalin version 3.1.0 again in [this commit](https://github.com/eclipse-emfcloud/emfcloud-modelserver/commit/6bcbd0efe7d1b0375c7aef0c7c9e9cce420e3f6e) to avoid the mocking problems that prevent the tests from running successful. In the long term, however, we should upgrade to a newer Javalin version. I therefore close this issue and created a [follow-up issue](https://github.com/eclipse-emfcloud/emfcloud-modelserver/issues/74) to monitor the situation to upgrade eventually.

PR #65, opened 2021-02-01
Stabilize broadcasting of incremental command updates

### LINKED ISSUE #66: Stabilize broadcasting of incremental command updates
In case of editing the name of an element (which also changes its `SemanticUriFragment`) via a `SetCommand`, the broadcasted incrementalUpdate (command) does not contain the correct element values in all cases.

Example:
`EClass` with name `"BrewingUnit"` -> set name `"BrewingUnitNew"`
-> SemanticUriFragment changes from `"//BrewingUnit"` to `"//BrewingUnitNew"`

If the command is encoded after executing the `SetCommand`, the broadcasted incrementalUpdate contains the updated element `"//BrewingUnitNew"` as owner and therefore the listeners have troubles consuming that update correctly.

As we already pre-encode commands in the case of Undo/Redo (to ensure proper command values for broadcasting to all listeners) we should stick to this behaviour also for executing other commands.

ndoschek:
Fixes #66 

### DIFF
--- a/bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/ModelController.java
+++ b/bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/ModelController.java
@@ -197,7 +197,7 @@ public class ModelController {
          boolean undoSuccess = modelRepository.undo(modeluri);
          if (undoSuccess) {
             ctx.json(JsonResponse.success("Successful undo."));
-            sessionController.broadcastUndoRedo(modeluri, encodings);
+            sessionController.modelChanged(modeluri, encodings);
             return;
          }
       }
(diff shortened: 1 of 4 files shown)


martin-fleck-at (changes requested):
Hi Nina, thank you very much for your change. I had a look at your changes and ran into a few issues:
- Test cases no longer compile, specifically ModelControllerTest and SessionControllerTest. Please ensure that they compile and run through.
- I seem to get updates multiple times: You can test this with the default client in the menu SuperBrewer3000.json > subscribeWithTimeout and then performing an edit operation. It may not happen the first time but if you wait for the timeout and redo those steps you should see multiple updates. You can do the same with SuperBrewer3000.coffee and its subscriptions. Could you look into that?

ndoschek:
Thank you for your review Martin and spotting my sloppy mistakes!
I fixed the test cases and they run through on my setup most of the time, occasional failures are known and caused by #63.
The multiple updates were caused due to a issue in [@eclipse-emfcloud/emfcloud-modelserver-theia](https://github.com/eclipse-emfcloud/emfcloud-modelserver-theia), i fixed it along with the current PR https://github.com/eclipse-emfcloud/emfcloud-modelserver-theia/pull/28.

martin-fleck-at (approved):
Thank you for your update, everything works as expected now! :+1:

PR #67, opened 2021-02-02
Fix #62: Improve and cleanup m2 and p2 build, add test execution

### LINKED ISSUE #62: Integrate tests into the CI

martin-fleck-at:
Maven
- Extract versions into properties
- Run test cases in m2 and p2 build
- Let pom files reflect directory structure
- Automatically download non-target platform jars into 'lib' directory

Target Platform
- Fix Jetty version for stability and faster target platform resolution

- Enhance Checkstyle
-- Allow method names with '_' in test files
-- Allow protected fields as we are a framework targeting extensibility
-- Remove FIXME and TODO comments

- Fix test cases
-- Downgrade to Javalin 3.1.0 to avoid mocking problem


releng/org.eclipse.emfcloud.modelserver.releng.p2/pom.xml
```diff
@@ -1,135 +1,134 @@
...
+                  <configuration>
+                     <executionEnvironment>JavaSE-11</executionEnvironment>
+                     <!-- IMPORTANT: DO NOT split the arg line -->
+                     <appArgLine>-application org.eclipse.ant.core.antRunner -buildfile packaging-p2composite.ant p2.composite.add -Dsite.label="Eclipse EMF.Cloud Model Server Nightly Updatesite"
+                        -Dproject.build.directory=${project.build.directory} -DunqualifiedVersion=${unqualifiedVersion} -DbuildQualifier=${buildQualifier} -Dsoftware.download.area=${rsync.local.dir}</appArgLine>
+                     <repositories>
+                        <repository>
+                           <id>2019-09</id>
+                           <layout>p2</layout>
+                           <url>http://download.eclipse.org/releases/2019-09</url>
```
ndoschek, line 82: I'm not sure, but should we use the `2020-09` repository here as well, as we use it in the target already?

ndoschek (commented):
Thank you very much for that great change Martin! Everything regarding the dependencies is much clearer to me now and it is great that tests are executed on every build from now on! :+1: 

I just left a few minor comments inline, and would have one request:
For some projects the preference that provides the code template for new Java files, holds an outdated header. This leads to an instant checkstyle warning on creating new Java files in those projects.
As you already optimized the checkstyle settings, it would be great if you could fix that in the course of this PR either with updating or removing this preference value. I do not have a strong opinion on that one, I'm fine with either one.
I took a look and it seems that all Java projects in the directories `bundles/` and `tests/` are affected, but please doublecheck.
The affected preference file is `/.settings/org.eclipse.jdt.ui.prefs` and the preference key is `org.eclipse.jdt.ui.text.custom_code_templates`.

releng/org.eclipse.emfcloud.modelserver.releng.p2/pom.xml
martin-fleck-at: Yes, you are right, great idea, I updated it as you suggested!

bundles/org.eclipse.emfcloud.modelserver.edit/src/org/eclipse/emfcloud/modelserver/edit/DefaultCommandCodec.java
```diff
@@ -104,7 +104,7 @@ public CCommand encode(final Command command) throws EncodingException {
             collectCommandEObjects(remove.getCollection(), result, false);
          }
       } else if (command instanceof SetCommand) {
-         // FIXME: Handle the UNSET value (needs model change)
+         // Still need to handle the UNSET value (needs model change)
```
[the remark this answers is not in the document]
martin-fleck-at: Hmm... I thought I saw some additional warnings pop up but you are right, we can actually keep them. Sorry for that and good catch! I restored them again!

martin-fleck-at:
@ndoschek Thank you for your thorough review! I fixed all the issues you mentioned and also incorporated your request (great suggestion!). With the latest changes everything should work again and no information should be lost.

martin-fleck-at:
@ndoschek I executed the build now several times with checkstyle issues and test failures and everything seems to report correctly in the CI. So you can have another go at this change if you have some time, thank you!

ndoschek (approved):
Thank you very much for your quick update on this PR!
It is very delightful now to create new classes and especially the improved CI build brings this project a big step further :+1:

Issue #72, opened 2021-02-09
Absolute full path as modelURI causes error


ndoschek:
Using an absolute full path like `file:///home/destination/file.coffee` leads to the following error:

```
9765 [qtp1598898814-26] DEBUG DefaultModelResourceManager  - Could not load resource with URI: file:///home/nina/Clients/OpenSource/emfcloud/emfcloud-modelserver/examples/org.eclipse.emfcloud.modelserver.example/.temp/workspace/SuperBrewer3000.coffee
[qtp1598898814-26] WARN io.javalin.Javalin - Uncaught exception
java.lang.NullPointerException
	at org.eclipse.emfcloud.modelserver.emf.common.DefaultModelResourceManager.removeResource(DefaultModelResourceManager.java:194)
	at org.eclipse.emfcloud.modelserver.emf.common.DefaultModelResourceManager.removeResourceSafe(DefaultModelResourceManager.java:186)
	at org.eclipse.emfcloud.modelserver.emf.common.DefaultModelResourceManager.loadResource(DefaultModelResourceManager.java:178)
	at org.eclipse.emfcloud.modelserver.emf.common.DefaultModelResourceManager.loadResource(DefaultModelResourceManager.java:165)
	at org.eclipse.emfcloud.modelserver.emf.common.ModelRepository.getModel(ModelRepository.java:59)
	at org.eclipse.emfcloud.modelserver.emf.common.ModelController.getOne(ModelController.java:106)
...
```

vhemery:
@ndoschek Sorry, but it looks like this issue fixed absolute path for linux but broke it for windows.
In org.eclipse.emfcloud.modelserver.emf.common.ModelServerRouting.adaptModelUri(String), `uri.toString()` used to work well for me. But now, `URI.createFileURI(uri.path()).toString();` gets rid of my device part of the path and gives me 404 errors. ( https://github.com/eclipse-emfcloud/emfcloud-modelserver/commit/b1eb06f5c36f0abe77f953aa030603449c17e95f )

Here is what my uri looks like and the various values of interest.
![image](https://user-images.githubusercontent.com/7920847/108193734-10ccab80-7116-11eb-9b65-083477c0fb43.png)
Maybe using `devicePath` or `toFileString` method ? Though I guess the issue would still be here because I have a blank authority...

I'm not sure what's the best fix here working for both linux and windows. Could you try and adapt this method to work with both URI patterns ?
Thanks in advance.

[...]

PR #80, opened 2021-02-17
Absolute full path as modelURI causes error #72

vhemery:
Fixing absolute file URI with device for windows.
Moving the adaptModelUri method to ModelResourceManager.


ndoschek (approved):
Thank for ensuring full paths work correctly on Windows as well! I tested the change on Linux and can verify everything works as expected. Making the modelUri handling customizable is a great benefit, thanks again! :tada:

Issue #94, opened 2021-04-01
Export org.eclipse.emfcloud.modelserver.edit.command.util package


ndoschek:
Export `org.eclipse.emfcloud.modelserver.edit.command.util` package to be able to access methods like `missingParameter`

PR #95, opened 2021-04-01
#94 Export org.eclipse.emfcloud.modelserver.edit.command.util package

ndoschek:


martin-fleck-at (approved):
Good catch, thank you Nina!

PR #97, opened 2021-04-02
#13, #74, #17: Fix some issues and update Javalin

### LINKED ISSUE #13: POST request to same resource should return 409
POST request triggers "Create new model" and should return 409 "Conflict" if tried again after model has been created successfully.

### LINKED ISSUE #17: Saving an unkown resource should not respond 'success'
When trying to save an unknown resource, the model server responds with success:
`{ "type": "success", "data": "Model 'file:/...' successfully saved" }`

This is not correct, it should return an error, just like for example the `DELETE` endpoint.
`{ "type": "error", "data": "Model 'file:/...' not found, cannot be deleted!" }`

### LINKED ISSUE #74: Upgrade to latest Javalin version as soon as mocking problems are resolved
This is a follow-up issue from https://github.com/eclipse-emfcloud/emfcloud-modelserver/issues/63 where it was reported that in newer Javalin versions mockito *sometimes* fails when mocking Javalin classes. There are several open issues related to this:
- https://github.com/tipsy/javalin/issues/979
- https://github.com/tipsy/javalin/issues/1098
- https://github.com/mockito/mockito/issues/1943

As soon as the issues are resolved we can upgrade.

martin-fleck-at:
Fixes #13: Trying to re-create an existing resource should return 409
Fixes #74: Upgrade to latest Javalin version
Fixes #17: Saving non-existing model should not return success


bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/DefaultModelController.java
```diff
@@ -72,7 +72,10 @@ public DefaultModelController(final ModelRepository modelRepository, final Sessi
    public void create(final Context ctx, final String modeluri) {
       Optional<EObject> root = readPayload(ctx);
       if (root.isEmpty()) {
-         badRequest(ctx, "Create new model failed.");
+         return;
+      }
+      if (this.modelRepository.getModel(modeluri).isPresent()) {
+         conflict(ctx, "Model already exists.");
```
ndoschek, line 78: This check seems to break the createModel request.
As the model is about to being created, `getModel` return an empty optional in that case.

ndoschek (changes requested):
Thank you very much for fixing these issues! :+1: 
Changes look good to me, except for one problem regarding the `createModel` request (please see inline).

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/DefaultModelController.java
martin-fleck-at: You are absolutely right, thank you! It seems that `getModel` already tries to load the model, so I changed the query to the safer `hasModel`

martin-fleck-at:
@ndoschek Thank you very much for that catch! I changed it and re-pushed the changes so we can Rebase & Merge them.

ndoschek (approved):

PR #98, opened 2021-04-02
Fix #87: Adapt command package namespace URI to better reflect source

### LINKED ISSUE #87: Command.ecore uses 'http://www.eclipsesource.com/schema/2019/modelserver/command' as Namespace URI
I wonder if the namespace URI for commands should be updated to something else.

martin-fleck-at:
- Use http://www.eclipse.org/emfcloud/modelserver/command as NS URI
- Adapt URI where necessary


ndoschek (approved):
LGTM :+1:

PR #106, opened 2021-05-04
#108 Disable macOS 10.15 runner for build-m2 action temporarily

ndoschek:
- Add ubuntu-latest to Github Actions runners
- Disable MacOS 10.15. (macOS-latest) temporarily for m2 build in GitHub Actions
- MacOS 11.0 is not available publicly yet (see https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ jobs:
       # do not cancel other builds if one fails
       fail-fast: false
       matrix:
-        os: [windows-2019, ubuntu-18.04, macos-10.15]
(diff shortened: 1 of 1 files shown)


martin-fleck-at (changes requested):
In theory the change looks good to me but is there a reason why we need two ubuntu versions to test? I'd suggest to either use latest for all or a fixed version for all but avoid having multiple windows or ubuntu versions mixed.

martin-fleck-at (approved):
Looks good to me, thank you Nina!

PR #113, opened 2021-06-02
Fix #111: Add error/failure logging and centralize package configs

### LINKED ISSUE #111: Ensure proper error logging for WebSocket SubscriptionListeners
By default, the `TypedSubscriptionListener<T>` overrides most interface methods with an empty body.
Especially in the case of `onFailure` and `onError` this should be improved to log the failure reason and its stack trace in any case.

martin-fleck-at:
Logging:
We want to to log errors and failures by default to prevent those
problems from being swallowed up since they may contain important
information about underlying issues.

EPackageConfigurations:
Centralize EPackageConfiguration management and apply it on the Java
model server client to ensure proper setup. We also ensure that the
ChangePackage is properly initialized on the client-side. We already use
the RecordingModelResourceManager as default on the server so we need
the ChangePackage to be initialized.


ndoschek (approved):
Thanks a lot for your change Martin! LGTM :+1: 
(PS: CI seems to have network issues again. I executed the builds locally and they went through fine in my case, so I will merge it)

PR #129, opened 2021-09-24
128: DefaultModelResourceManager deletes files it can't load

CamilleLetavernier:
- Remove the resource from the resource set, without actually deleting it
- Report the loading issue as an error, rather than debug message

I've also included two minor fixes that have been bugging me for some time :slightly_smiling_face: :

- Remove invalid package export from modelserver.emf Manifest
- Fix NPE when reporting an exception that doesn't have a message


CamilleLetavernier:
Actually, changing the log severity is not such a good idea: this method is invoked on all files in the workspace, including the ones that aren't EMF Models. So, we don't want to log errors for these. 

I'm not sure what's the best way to distinguish this first step of "Try to load all files and see what happens" (where we don't want to log issues) and the "Please load this model, I want to edit it" step, where errors should actually be logged. Any suggestions?

ndoschek:
Thanks Camille for addressing those issues!
Maybe we should rethink the functionality of `loadSourceResources` a little bit as well. As we are aware of the registered `EPackageConfiguration`s, wouldn't it be enough to only scan and load resources that are meant to be managed by the Model Server (i.e. filter the workspace contents for those registerted file types)?
Also, at a later point in time, the `loadResource` method will not be used for unknown resources anyway and therefore changing the logging severity would be okay I would say. But if the resource becomes erroneous during editing, I'm not sure if removing the resource is the best solution in the first place, what do you think?

Maybe @martin-fleck-at has also some ideas regarding these issues.

CamilleLetavernier:
> Maybe we should rethink the functionality of loadSourceResources a little bit as well. As we are aware of the registered EPackageConfigurations, wouldn't it be enough to only scan and load resources that are meant to be managed by the Model Server (i.e. filter the workspace contents for those registerted file types)?

That's a bit risky. For example, if EPackages are registered by extension point (using the `org.eclipse.emf.ecore.extension_parser` extension point), then the EPackageConfiguration doesn't need to specify file extensions anymore (Actually, specifying a file extension would override the declaration from the extension point, which might lead to unexpected issues).

We could use EMF-based content types, but they aren't always specified, either.

In general, do we even need to pre-load all models? I think we have an endpoint to get all models; but we only use it in some specific use cases, I believe. So maybe the "Load all models" action could remain, but only when we explicitly call this getModels endpoint? And then, trying to load all resources wouldn't be too much of an issue (Although that probably wouldn't scale on real-life workspaces, which would likely contain tons of code, including generated and compiled files).

So, it means we need a strategy to identify which files are models, and avoid loading all unrelated files. But I'm not sure what's the best way to do that. I'm not even sure there is a universal solution for that. Maybe the common Model Server could provide the endpoint, and a generic, naive strategy (e.g. based on the file extension, as you suggested), but keep it very simple and very easy to override?

> But if the resource becomes erroneous during editing, I'm not sure if removing the resource is the best solution in the first place, what do you think?

No, we should probably never have to remove the resource from the resource set (Except during explicit "unload/close" requests). If the resource is invalid, there is no point in trying to reload it later on, unless the file has changed on the disk. But then, we should have a specific listener & strategy for that: if the file changes, we need to decide if we want to reload it, or keep the current version in memory. But I think we already have a different story for that.

martin-fleck-at:
@CamilleLetavernier Having a look at the PR and the discussion I do agree with you that the model server should never just delete a file if there is an exception.

The model server should manage the resource set properly so we can argue whether or not it makes sense to keep a broken resource in the resource set. The same goes for the initial scanning of the resources in the workspace which makes the exception handling now a bit more difficult. I would hope that both problems are easier to solve once we have proper file watching support as then we know when to reload a resource in the resource set and do not have the risk of having the broken resource there forever. However, as you mentioned this is a bit of a different discussion.

For now I believe your change is already providing some value. Since we do not have any file watching yet, I believe it might be nicer to extract the whole handling of unloadable resources into a separate method that the user can override. By default we should probably log the problem and remove the resource from the resource set as we do now. This way we can also get rid of the `removeUnloadableResources` flag and (for the time being) introduce a `isInitializing` flag instead. This flag would help us identifying whether we are in the workspace initialization phase where we know we may encounter unloadable resources or we are in the explicit usage phase where resources should be loadable.

What do you think?

CamilleLetavernier:
Sounds good to me. I'll update the patch

CamilleLetavernier:
I just noticed that there is also a `removeErroneousResources` method, which mentions: `any resources loaded with errors are probably not resources in the first place`.

While that's mostly true in our current web-based use cases (where most errors are caused by files that aren't EMF Models), it is actually really common to load "broken" EMF Models in practice. The most typical issue is unresolved proxies ("resource not found"), which won't prevent from loading the resource/model, but may still be reported as a loading error (Although that usually happens later, during lazy-resolution of these unresolved proxies, and not immediately after loading the resource)

Files that aren't resources will typically be empty (Resource.getContents() is empty), whereas files with non-critical errors may contain error, but will still contain some objects. But we really need a more reliable mechanism to identify which files are resources, and which ones are not.

CamilleLetavernier:
I realized that changing the behavior of removeResource() was not correct: this is actually a public API used by the ModelRepository/ModelController to physically delete a model, so we shouldn't change its behavior.

I've amended the commit with a few changes:

- Rename `removeResource` to `deleteResource`, for consistency with the endpoint (`delete` Request), and to avoid further confusion. `deleteResource` keeps its behavior of physically deleting the resource/file
- Add a new `removeResource` (internal) method, to unload a Resource and remove it from the ResourceSet
- Remove the `removeUnloadableResources` parameter from the `loadResource` method. We now only have one method for `loadResource` (vs one with a parameter, and one without it)
- Add an `isInitilializing` flag to the Resource Manager
- Add a protected `handleLoadError` method to handle resource-load exceptions. The default behavior is to log an error (Or debug during initialization), and to remove the resource from the resource set

CamilleLetavernier:
One more change: I've updated the tests. The "testLoadModelFromInvalidModelUri" was expecting a NullPointerException, because we don't create ResourceSets for invalid resources; so the removeResource() method was causing an NPE (Which wasn't caught by removeResourceSafe, because that method only caught IOExceptions).

With the new methods structure, removeResourceSafe no longer exists, and removeResource already checks for Null-ResourceSets, so we won't have exceptions during resource loading (Everything should be safely handled in handleLoadError()). I've updated the test to expect an Optional.empty() result rather than a NPE.

CamilleLetavernier:
One final (?) change to fix Checkstyle issues

PR #134, opened 2021-11-04
#115 : Introduce file watching concept

### LINKED ISSUE #115: Introduce file watching concept
At the moment, if resources are manipulated without direct Model Server requests (regardless of which clients are using ite.g. directly via the file system or via Theia's default editors), the Model Server is not aware of such file manipulations.
We should introduce a file/directory watching concept directly on the Model Server and have a strategy for reconciling (Essentially: either keep the in-memory model and ignore file-system changes, or reload from file system).

(Issue was initially created in `emfcloud-modelserver-theia`: https://github.com/eclipse-emfcloud/emfcloud-modelserver-theia/issues/50)

vhemery:
Adding ModelWatchers, which must invoked by ModelResourceManager after
loading a resource.
On a divergence, a ModelWatcher must ask the active ReconcilingStrategy
to solve the issue. (2 are contributed)
The default FileModelWatcher implementation is contributed, listening at
file changes, but other implementations are possible for different
protocols.

Fixes #115

### DIFF
--- a/bundles/org.eclipse.emfcloud.modelserver.emf/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.emfcloud.modelserver.emf/META-INF/MANIFEST.MF
@@ -23,6 +23,7 @@ Require-Bundle: com.fasterxml.jackson.core.jackson-core;bundle-version="[2.9.0,3
 Export-Package: org.eclipse.emfcloud.modelserver.emf.common,
  org.eclipse.emfcloud.modelserver.emf.common.codecs,
  org.eclipse.emfcloud.modelserver.emf.common.util,
+ org.eclipse.emfcloud.modelserver.emf.common.watchers,
  org.eclipse.emfcloud.modelserver.emf.configuration,
  org.eclipse.emfcloud.modelserver.emf.di,
  org.eclipse.emfcloud.modelserver.emf.launch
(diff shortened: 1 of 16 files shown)


vhemery:
I must still update the tests with correct Guice bindings...
Will push an update soon...

vhemery:
This fail is not on me...
Jetty 9.4.43.v20210629 has disappeared during the night. I think we should update to 9.4.44, then retrigger the build...

vhemery:
It works on my machine ¯\_(ツ)_/¯

I think this failure is related to the fact that reconciliation occurs in a separate daemon thread, which may not have a priority big enough on the integration platform (relatively to build threads).

I will try and increase the thread priority only for the test...

martin-fleck-at:
@vhemery It seems that the build runs through but there are some checkstyle issues: https://ci.eclipse.org/emfcloud/job/eclipse-emfcloud/job/emfcloud-modelserver/job/PR-134/10/checkstyle/

vhemery:
Finally nailed it... I was missing the first change notification because of late thread start.

> 
> 
> @vhemery It seems that the build runs through but there are some checkstyle issues: https://ci.eclipse.org/emfcloud/job/eclipse-emfcloud/job/emfcloud-modelserver/job/PR-134/10/checkstyle/

Thanks that was exactly what I was looking for...

vhemery:
@martin-fleck-at  @ndoschek 
It would be very nice if we could get this merged by end of the week. (pretty pleaaaase)
Of course, the target platform part conflicts with Camille's PR (who does the same fix).
In case you want to merge #136 before, I can easily fix this PR with a rebase, since the target platform part is in a separate commit.

Please, feel free to ask any question in case you need extra information on this PR.

martin-fleck-at:
@vhemery Hello Vincent! We are sorry about the delay but something else always came up. However, we can merge it by the end of this week for sure! So stay tuned ;-)

Issue #135, opened 2021-11-08
The ModelServer is not thread-safe


CamilleLetavernier:
When executing quick actions, the model server often throws unexpected exceptions, mostly in EMF APIs (For example, IndexOutOfBounds when serializing EMF-to-JSon, or when iterating over an EMF List, ...). These errors often indicate concurrent access to the model from different threads, which is not supported by EMF out of the box.

When integrating with GLSP, the issue becomes worse, as ModelServer-GLSP may also handle model updates and replies from the Model Server in the wrong thread. This causes a lot of instabilities, from which it is often not possible to recover (The only solution with a corrupt EMF Model is to restart the application/server)

PR #136, opened 2021-11-08
#135 Introduce Thread-Safe ModelController

CamilleLetavernier:
Introduce a new ModelController, that handles all incoming requests in the same thread, and delegates the actual behavior to an existing implementation of a ModelController.

This is a simple (and naive) way of ensuring thread-safety. For more complex use cases, we may want to use per-resource set threads instead, or even rely on EMF Transactions (To support single-write, concurrent-read access); but having basic support is probably good enough for now.

This approach is probably less disruptive than modifying the default ModelController, as we should be able to swap the delegate to any existing implementation of a ModelController. It **should** be transparent for existing modules as well (Provided they extend the `DefaultModelServerModule`), but clients should still double-check if the 2 ModelControllers (ThreadSafe & Delegate) are properly bound.

### DIFF
--- a/bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/SingleThreadModelController.java
+++ b/bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/SingleThreadModelController.java
@@ -1,0 +1,225 @@
+/********************************************************************************
+ * Copyright (c) 2021 EclipseSource and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * https://www.eclipse.org/legal/epl-2.0, or the MIT License which is
+ * available at https://opensource.org/licenses/MIT.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR MIT
+ ********************************************************************************/
+package org.eclipse.emfcloud.modelserver.emf.common;
+
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.FutureTask;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/SingleThreadModelController.java
```diff
@@ -0,0 +1,225 @@
...
+   protected final ModelController delegate;
+
+   protected final Thread thread;
+
+   protected final BlockingQueue<Runnable> actionsQueue = new ArrayBlockingQueue<>(100, true);
+
+   protected final String name;
+
+   @Inject
+   public SingleThreadModelController(final @Named("ModelControllerDelegate") ModelController delegate) {
```
ndoschek, line 47: We should use a constant for the `"ModelControllerDelegate"`, e.g. as for the preferred format annotation in https://github.com/eclipse-emfcloud/emfcloud-modelserver/blob/873c12dbfca2423fbd5a0d3e6fdcf10fb78a58b2/bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/codecs/CodecsManager.java#L32
https://github.com/eclipse-emfcloud/emfcloud-modelserver/blob/873c12dbfca2423fbd5a0d3e6fdcf10fb78a58b2/bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/codecs/DICodecsManager.java#L38

ndoschek (changes requested):
Thanks a lot Camille, changes look great to me! :tada: 
I found a minor issue regarding the named annotation.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/SingleThreadModelController.java
CamilleLetavernier: Good point. Done :+1:

ndoschek (approved):
Thanks for the quick update!

PR #137, opened 2021-11-17
Fix M2 build and add secondary license to LICENSE file

ndoschek:


martin-fleck-at (approved):
Change looks good to me, thank you Nina!

PR #142, opened 2021-11-29
#47 Workspace configuration not working in Windows

tortmayr:
Updating the configuration via the `server/configure` endpoint does not work under Windows when used in combination with Modelserver Theia.
An absolute Windows path typically contains a driver letter followed by a : (e.g c://my-workspace). So when passing absolute paths as URL parameters the `:` need to be encoded.
This means we have to make sure that they are properly decoded on the server side.
The easiest common point to this this the `DefaultUriHelper` class.

Fixes #47


ndoschek (approved):
Thanks a lot Tobias, change looks good to me and works great! :+1: 
I added an additional test case for the decoded url case.

Issue #150, opened 2022-01-17
ModelServer API v2


sdirix:
The current ModelServer API often directly serializes EMF Objects to JSON and directly exposed the EMF command stack and change notifications. Going forward it would be nice to ensure the technology independence of the API, in particular removing nomenclature, syntax, and structures that are biased to the Eclipse Modeling Framework (EMF). This would help to remove barriers to implementation of client applications and server components in Javascript and other web technologies.

We therefore suggest developing a version 2 of the API which is more convenient for Javascript clients but also keeps the current benefits for Java clients.

martin-fleck-at:
Hi Stefan,

Great idea! I agree that there are a few areas that definitely need improving (e.g., change notifications as discussed in https://github.com/eclipse-emfcloud/emfcloud-modelserver/issues/79) and I like the idea of a version 2. It'd definitely help us get a cleaner API with all the experience and use cases we have collected during development of v1 without breaking existing clients. I also think removing the EMF nomenclature from the API is a good idea as it reveals more about the implementation than about the functionality. So I think this is a great chance for us, as long as Java clients are still fully supported and provide the benefits we currently have using EMF more or less directly.

With regards to v1, the most important part for me is that it still works as it does now to avoid forcing people to update. Any new functionality, however, does not need to be back ported in my opinion. Also, we should consider a clear and detailed documentation for v2 and provide a guide that details how adopters can migrate from v1 to v2. If we have all that, I don't think there should be any problems with providing a v2.

PR #151, opened 2022-01-18
150: ModelServer API v2

CamilleLetavernier:
Introduce a new version for the ModelServer API (V2). This version removes some of the EMFisms from the API:

- All Json Model Elements now consistently include $type and $id (In v1, we had "eClass" and no specific id attribute). References still use $ref.
- Edition can be done via EMF CCommands (as in V1) or Json Patch (New in V2). The patch or command needs to be wrapped in an object specifying the format used:
```
{
    "type": "modelserver.emfcommand"
    "data": { "...v1-command..." }
}
```
At the moment, supported types are "modelserver.emfcommand" (V1 CCommand) and "modelserver.jsonpatch" (V2 Json Patch).
- A new endpoint has been added to support transactions. This will complement (or even replace) Custom Commands. Clients may start a transaction, and then execute multiple sequential commands. When the transaction is closed, all subscribed clients will be notified.
- Subscription notifications will be sent as a Json Patch (Instead of EMF ChangeDescription) for V2 clients.


CamilleLetavernier:
Note: this is a draft PR. Some pieces are still missing (especially, we need to improve the support for Json Patch operations. Currently, only 'replace' is implemented).

The ModelServer should still support the V1 endpoints, but using both V1 and V2 endpoints at the same time (for different clients) may cause unexpected issues (Especially regarding Command execution/notifications).

CamilleLetavernier:
The PR includes some library for manipulating Json Patch on the server. The library is `json-patch` (`java-json-tools`), and some dependencies. I had to repackage them, because even though the original library provides some OSGi metadata, they seem to be incomplete or invaliid (Especially, not all packages are exported, so we can't consume it).

The libraries are included in the `lib` folder, and are available under a dual license (LGLP v3 or Apache V2). So, license shouldn't be an issue (But we'll need to fill a CQ).

CamilleLetavernier:
Added implementation & (integration) tests for Add (Create) and Remove operations. Move and Test are still unsupported.

CamilleLetavernier:
The new commit attaches the Json Patch node to the result, after executing a Command/Patch operation. The "data" field is modified to support both a message and a patch.

V1 Result:
```
{
    "type": "success",
    "data": "Model 'file:/.../SuperBrewer3000.coffee' successfully updated"
}
```
V2 Result:
```
{
    "type": "success",
    "data": {
        "message": "Model 'file:/.../SuperBrewer3000.coffee' successfully updated",
        "patch": [
            {
                "op": "add",
                "path": "/workflows/-",
                "value": {
                    "$id": "//@workflows.7",
                    "name": "New WF"
                }
            }
        ]
    }
}
```

(Note: the same thing needs to be done for 'undo' and 'redo' endpoints; but at the moment these endpoints are shared between V1 and V2, so they'll need to be split)

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/DefaultSessionController.java
```diff
@@ -197,39 +225,82 @@ protected void broadcastIncrementalUpdates(final String modeluri, final CCommand
       getOpenSessions(modeluri).forEach(session -> broadcastIncrementalUpdate(session, updates));
    }
 
+   protected void broadcastIncrementalUpdatesV2(final String modeluri, final JsonNode jsonPatch) {
+      // TODO Properly specify the message format. Currently reusing the same signature as V1,
```
eneufeld, line 229: Pull out Todo into an issue?

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/ModelController.java
```diff
@@ -39,8 +42,31 @@
...
+    * a Json Object with a "type" attribute, where type may be
+    * either {@link ModelServerPathsV2#JSON_PATCH} or
+    * {@link ModelServerPathsV2#EMF_COMMAND}, and a "data" attribute
+    * representing a Command or a Patch, in a format that corresponds
+    * to the specified type.
+    *
+    * @param ctx
+    * @param modelURI
+    */
+   default void executeCommandV2(final Context ctx, final String modelURI) {
```
eneufeld, line 66: why define it as default and throw an exception?

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/ModelServerRoutingV2.java
```diff
@@ -0,0 +1,101 @@
...
+   protected void openTransaction(final WsHandler wsHandler) {
+      wsHandler.onConnect(transactionController::onOpen);
+      wsHandler.onClose(transactionController::onClose);
+      wsHandler.onError(transactionController::onError);
+      wsHandler.onMessage(transactionController::onMessage);
+   }
+
+   @Override
+   protected void executeCommand(final Context ctx) {
+      // TODO: this was rewired from PATCH /edit to PATCH /models
```
eneufeld, line 95: Create issue for this?

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/codecs/JsonCodecV2.java
```diff
@@ -0,0 +1,66 @@
...
+      mapper.configure(SerializationFeature.INDENT_OUTPUT, true);
+      mapper.setDateFormat(dateFormat);
+      mapper.setTimeZone(TimeZone.getDefault());
+      mapper.registerModule(new EMFModuleV2());
+      return mapper;
+   }
+
+   @Override
+   public JsonNode encode(final EObject obj) throws EncodingException {
+      // FIXME: The super implementation creates a JsonResource, which doesn't (seem to) support
```
eneufeld, line 57: In all custom tools, eg ecore-glsp or coffee-editor we are doing exactly the same and enable the ids. 
I'm wondering whether we should rather enable ids for the default JSONCodec and thus get rid of the JsonCodecV2 and the whole json2 resource identifier? @ndoschek @martin-fleck-at what do you think?

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/util/ContextRequest.java
```diff
@@ -145,4 +146,34 @@ public static boolean isMessageType(final WsMessageContext ctx, final String typ
...
+      try {
+         JsonNode json = JavalinJackson.getObjectMapper().readTree(ctx.message());
+         if (!json.has(JsonResponseMember.TYPE)) {
+            error(ctx, "Missing message type");
+            return Optional.empty();
+         }
+
+         JsonNode jsonTypeNode = json.get(JsonResponseMember.TYPE);
+         String jsonType = !jsonTypeNode.asText().isEmpty() ? jsonTypeNode.asText() : jsonTypeNode.toString();
+         if (jsonType.equals("{}")) {
```
eneufeld, line 160: wasn't there a json util which checked for empty string?

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/util/JsonPatchHelper.java
```diff
@@ -0,0 +1,152 @@
...
+
+   @Override
+   protected TransactionalEditingDomain getEditingDomain(final EObject eObject) {
+      ResourceSet resourceSet = eObject.eResource().getResourceSet();
+      return modelManager.getEditingDomain(resourceSet);
+   }
+
+   @Override
+   protected Resource getResource(final String modelURI, final URI resourceURI) {
+      // FIXME We should make sure that modelURI and resourceURI are part of the same
```
eneufeld, line 57: create an issue?

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/util/JsonPatchHelper.java
```diff
@@ -0,0 +1,152 @@
...
+    *         A JsonPatch representing the differences between oldModel and newModel
+    * @throws EncodingException
+    */
+   public JsonNode diffModel(final JsonNode oldModel, final JsonNode newModel)
+      throws EncodingException {
+      // TODO Support multiple resource patches
+      return JsonDiff.asJson(oldModel, newModel);
+   }
+
+   // public JsonNode getCurrentModel(final String modeluri, final Optional<ArrayNode> jsonPatch)
```
eneufeld, line 83: remove commented out code?

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/util/JsonPatchHelper.java
```diff
@@ -0,0 +1,152 @@
...
+    *                    The old model, before applying changes
+    * @param newModel
+    *                    The new model, after applying changes
+    * @return
+    *         A JsonPatch representing the differences between oldModel and newModel
+    * @throws EncodingException
+    */
+   public JsonNode diffModel(final JsonNode oldModel, final JsonNode newModel)
+      throws EncodingException {
+      // TODO Support multiple resource patches
```
eneufeld, line 79: new issue?

pom.xml
```diff
@@ -297,7 +298,7 @@
 									<groupId>org.eclipse.emfcloud.modelserver</groupId>
 									<artifactId>org.eclipse.emfcloud.modelserver.parent</artifactId>
 									<classifier>releng/org.eclipse.emfcloud.modelserver.releng.target/targetplatform</classifier>
-									<version>${project.version}</version>
+									<version>0.7.0-SNAPSHOT</version>
```
eneufeld, line 301: why not use the project version anymore?

tests/org.eclipse.emfcloud.modelserver.edit.tests/build.properties
```diff
@@ -1,5 +1,5 @@
 source.. = src/
-output.. = bin/
+output.. = target/classes/
```
eneufeld, line 2: can we leave this on the eclipse standard and keep `bin`?

eneufeld (commented):
I think this looks great thank you. 
I have mainly comments about creating follow up issues and some possible generic improvements.

bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/client/AbstractModelServerClient.java
```diff
@@ -0,0 +1,807 @@
...
+import java.util.Set;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+
+import org.apache.log4j.Logger;
+import org.eclipse.emf.ecore.EObject;
+import org.eclipse.emfcloud.modelserver.common.ModelServerPathParameters;
+import org.eclipse.emfcloud.modelserver.common.ModelServerPathParametersV1;
+import org.eclipse.emfcloud.modelserver.common.ModelServerPathsV1;
```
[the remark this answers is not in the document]
CamilleLetavernier: Could be, but since V2 is still a moving target, it's hard to say what's common and what might still change. Having duplicate classes is a bit easier to work with. But I don't mind either way; I think we're close to the final API now (And parameter constants are less likely to change anyway)

bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/client/AbstractModelServerClient.java
```diff
@@ -0,0 +1,807 @@
...
+
+   /**
+    * Test whether format is supported. You may override this method if your model server has specific format and codecs
+    * configuration.
+    *
+    * @param format the format to test.
+    * @return true when supported.
+    */
+   protected boolean isSupportedFormat(final String format) {
+      return DEFAULT_SUPPORTED_FORMATS.contains(format);
```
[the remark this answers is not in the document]
CamilleLetavernier: This is supported formats for Codecs, to encode/decode the Model or the Commands. I don't think EPackages make any difference here.

We expect either XMI or Json/JsonV2, and that's it

bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/client/AbstractModelServerClient.java
```diff
@@ -0,0 +1,807 @@
...
+      return value.orElseThrow(() -> new RuntimeException("Could not parse 'data' field"));
+   }
+
+   public String encode(final EObject eObject) {
+      return encode(eObject, getDefaultFormat());
+   }
+
+   public String encode(final EObject eObject, final String format) {
+      try {
+         if (format.equals(ModelServerPathParametersV1.FORMAT_XMI)) {
```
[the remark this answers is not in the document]
CamilleLetavernier: I'm not sure it really makes a difference, since we only expect XMI or Json (We don't even check registered Codecs here). But we should probably add support for decoding Json-V2 objects (Currently, we only use Json-V2 for encoding Models)

pom.xml
CamilleLetavernier: Because of the introduction of external libraries, which are not versioned as "0.7.0", the target platform couldn't resolve. The alternative would be to override the target platform definition for these bundles that use a different version, but I think having a single target platform is more consistent (rather than assuming we'll have a target platform version per bundle version).

tests/org.eclipse.emfcloud.modelserver.edit.tests/build.properties
CamilleLetavernier: I don't mind either way, as long as .classpath and build.properties are consistent :) all other plug-ins currently use target/classes/

bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/client/AbstractModelServerClient.java
```diff
@@ -0,0 +1,807 @@
...
+
+   public CompletableFuture<Response<Boolean>> configure(final ServerConfiguration configuration) {
+
+      ObjectNode config = Json.object(
+         Json.prop("workspaceRoot", Json.text(configuration.getWorkspaceRoot())),
+         Json.prop("uiSchemaFolder", Json.text(configuration.getUiSchemaFolder())));
+
+      final Request request = new Request.Builder()
+         .url(makeUrl(ModelServerPathsV1.SERVER_CONFIGURE))
+         .put(RequestBody.create(config.toString(), MediaType.parse("application/json")))
```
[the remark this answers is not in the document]
CamilleLetavernier: MediaType is not one of our classes, but having a constant would make sense :+1:

tests/org.eclipse.emfcloud.modelserver.edit.tests/build.properties
eneufeld: me neither, as long as it is consistent

eneufeld:
> The PR includes some library for manipulating Json Patch on the server. The library is `json-patch` (`java-json-tools`), and some dependencies. I had to repackage them, because even though the original library provides some OSGi metadata, they seem to be incomplete or invaliid (Especially, not all packages are exported, so we can't consume it).
> 
> The libraries are included in the `lib` folder, and are available under a dual license (LGLP v3 or Apache V2). So, license shouldn't be an issue (But we'll need to fill a CQ).

We could use the maven support for target files, but we would need to get rid of the tpd files then ... . I think adding them to the lib folder is fine.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/util/ContextRequest.java
ndoschek: `JsonNode` offers an `isEmpty` method itself afaik.
```suggestion
         if (jsonType.isEmpty()) {
```

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/codecs/JsonCodecV2.java
ndoschek: Yes that sounds good to me and would simplify things already lot for any adopters.

ndoschek (commented):
Thanks a lot for this valuable contribution! I have some rather general remarks/questions on the PR, please see below.

- Directory `libs`
  - Why is it necessary to store them in a separate directory? Could we include them in the existing `bundles/org.eclipse.emfcloud.modelserver.lib`?
  - If they should remain as separate projects, please update their project settings, project files and classpaths according to the [README](https://github.com/eclipse-emfcloud/emfcloud-modelserver#create-new-project) as they cause file changes on import.

- General remarks/questions
1. Please take a look at the new warnings in the workspace and remove them if possible
2. For newly added files, the header should only contain `2022` and not a year range. Also, some of those newly added files have a `EclipseSource` header and some a `STMicroelectronics` header. Should they be unified?
3. Why exactly do we need the `targetplatform-dev.target`? Would the additional dependencies hurt too much in the main targetplatform? However, it makes the development setup a little bit more complex imo, so if it needs to stay, we should at least mention that in the README.
4. Will this story include API documentation as well (@eneufeld/@sdirix)? In general, the current documentation should be improved anyhow. It would be great to offer API documentation for example via [OpenAPI](https://javalin.io/plugins/openapi#getting-started) and V2 would be definitely a great starting point to do so.

CamilleLetavernier:
> Why is it necessary to store them in a separate directory? Could we include them in the existing bundles/org.eclipse.emfcloud.modelserver.lib?

I didn't realize we already had a lib bundle. The approach I took is probably closer to the typical Orbit approach (keeping the original names), but adding all libs to the same bundle would definitely work, too. And it would be more consistent with the current ModelServer approach.

> For newly added files, the header should only contain 2022 and not a year range. Also, some of those newly added files have a EclipseSource header and some a STMicroelectronics header. Should they be unified?

Actually, these files are not "newly added", they're forked/duplicated from existing classes. So, I kept the original Copyright in these cases. I double-checked yesterday, and all copyrights should be consistent: new files have the ST-2022 Copyright; existing and forked files have the ES-xxxx-2022 Copyright.

> Why exactly do we need the targetplatform-dev.target? Would the additional dependencies hurt too much in the main targetplatform? However, it makes the development setup a little bit more complex imo, so if it needs to stay, we should at least mention that in the README.

This looks like a mistake. I played with target platforms a bit during development, but the current -dev platform is definitely not necessary. I'll remove it.

> Will this story include API documentation as well (@eneufeld/@sdirix)?

That definitely makes sense. However, I think it's still a bit early for that, as some APIs and use cases are not clearly defined yet (Especially URI handling, which Json Patches we support, how responses/notifications are sent, and how to handle multi-resources use cases). But we should probably start working on an initial version soon.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/util/ContextRequest.java
CamilleLetavernier: The pattern seems similar to what already existed in this class. As far as I understand, JsonNode.isEmpty() is not strictly equivalent, because it would match both `[]` and `{}`, as well as `""`.

But I think 'type' is expected to be a string anyway, so anything other than non-empty text should be invalid. If I read the code correctly, we currently accept e.g. `"type": {"a": true, "b": "value"}` as a valid type (Parsed a a string: `{"a": true, "b": "value"}`), which doesn't make a lot of sense.

Using `jsonTypeNode.isTextual() && !jsonTypeNode.isEmpty()` would IMO be more consistent.

@cdamus WDYT?

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/util/JsonPatchHelper.java
CamilleLetavernier: Issue: https://github.com/eclipse-emfcloud/emfcloud-modelserver/issues/159

bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/internal/client/EditingContextImpl.java
```diff
@@ -11,24 +11,28 @@
...
+
 import okhttp3.Response;
 import okhttp3.WebSocket;
 import okhttp3.WebSocketListener;
 
 /**
  * Implementation of the client's editing context.
  */
-public class EditingContextImpl extends WebSocketListener implements EditingContext {
+public class EditingContextImpl<CLIENT> extends WebSocketListener implements EditingContext {
```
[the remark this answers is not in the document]
CamilleLetavernier: I'm not sure. We only use the EditingContext with AbstractModelServerClient, but the EditingContext itself doesn't care which kind of client it is. As long as you give it a valid CCommand encoder, it can handle any client class.

@cdamus What do you think?

bundles/org.eclipse.emfcloud.modelserver.common/src/org/eclipse/emfcloud/modelserver/common/patch/AbstractJsonPatchHelper.java
```diff
@@ -0,0 +1,392 @@
...
+      if (jsonPatch.isEmpty()) {
+         return IdentityCommand.INSTANCE;
+      } else if (jsonPatch.size() == 1) {
+         return getCommand(modelURI, resourceSet, jsonPatch.get(0));
+      } else {
+         // CompoundCommands and multi-operation Json Patches
+         // are not equivalent. Json Patch operations should be applied
+         // sequentially. For this reason, we need to create a command, execute
+         // it, and then create the next command. We use LazyCompoundCommand for this.
+         SimpleLazyCompoundCommand command = new SimpleLazyCompoundCommand("Json Patch"); // TODO We should add a label
```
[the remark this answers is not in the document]
CamilleLetavernier: Created #160

bundles/org.eclipse.emfcloud.modelserver.common/src/org/eclipse/emfcloud/modelserver/common/patch/AbstractJsonPatchHelper.java
```diff
@@ -0,0 +1,392 @@
...
+         // References
+         JsonNode refNode = value.get(JsonConstants.REF_ATTR);
+         JsonNode idNode = value.get(JsonConstants.ID_ATTR);
+         if (refNode == null && idNode == null) {
+            throw new JsonPatchException("Reference values should include a $ref or an $id attribute");
+         }
+         String objectId = refNode == null ? idNode.asText() : refNode.asText();
+         return getEObject(modelURI, URI.createURI(objectId));
+      }
+      // TODO Support datatypes?
```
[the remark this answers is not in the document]
CamilleLetavernier: Created #162

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/DefaultSessionController.java
CamilleLetavernier: If possible, this should be handled in the current PR. Let's come back to this once support for Json Patches and Notifications is ready.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/ModelController.java
CamilleLetavernier: The idea is that current custom V1 implementations won't have to change anything (but they won't get any support for V2 either). So they can keep running without any change, and if they want to support V2, they'll have to implement the new methods.

If clients try to reach the V2 API for these V1 implementations, they'll receive the exception.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/ModelServerRoutingV2.java
CamilleLetavernier: We should probably consider this is the context of this PR, as this might cause unexpected behavior for V1 (Breaking legacy behavior).

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/codecs/JsonCodecV2.java
CamilleLetavernier: JsonV1 and JsonV2 have other differences, especially the way eClass and id are serialized. In V1, we have `{eClass: "EMF Type"}`, whereas in V2 we have `{$type: "EMF Type", $id: "EMF Fragment"}

So even if we improve the support for IDs in V1, we still need to distinguish the codecs.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/util/JsonPatchHelper.java
CamilleLetavernier: This should probably be handled with #159

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/util/ContextRequest.java
cdamus: Yes, I prefer Camille's last suggestion as it's more consistent with the abstract meaning. WBN not to compare things to the concrete JSON syntax.

bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/internal/client/EditingContextImpl.java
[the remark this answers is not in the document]
cdamus: I try to keep generic signatures as wide open as possible for clients and as narrow as possible for providers. I don't think introducing a generic type here helps the implementation in any way and it wouldn't mean anything to a user/client of the API.

eneufeld (commented):
I have some more minor comments and questions left

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/util/ContextRequest.java
```diff
@@ -145,4 +146,34 @@ public static boolean isMessageType(final WsMessageContext ctx, final String typ
...
+            error(ctx, "Missing message type");
+            return Optional.empty();
+         }
+
+         JsonNode jsonTypeNode = json.get(JsonResponseMember.TYPE);
+         String jsonType = jsonTypeNode.isTextual() && !jsonTypeNode.asText().isEmpty() ? jsonTypeNode.asText() : null;
+         if (jsonType == null) {
+            error(ctx, "Invalid message type");
+            return Optional.empty();
+         }
```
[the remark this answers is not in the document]
CamilleLetavernier: Checkstyle agrees :) 

> Cyclomatic Complexity is 8 (max allowed is 7).

bundles/org.eclipse.emfcloud.modelserver.common/src/org/eclipse/emfcloud/modelserver/common/ModelServerPathsV2.java
```diff
@@ -0,0 +1,62 @@
...
+   String VALIDATION = "validation"; // accepts query parameter "modeluri"
+   String VALIDATION_CONSTRAINTS = "validation/constraints"; // accepts the query parameter "modeluri"
+
+   String TYPE_SCHEMA = "typeschema"; // accepts query parameter "modeluri"
+   String UI_SCHEMA = "uischema"; // accepts query parameter "schemaname"
+
+   String SERVER_CONFIGURE = "server/configure";
+   String SERVER_PING = "server/ping";
+
+   // TODO(v2): New paths for new v2 APIs
```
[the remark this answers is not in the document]
CamilleLetavernier: Already removed in the previous commit

tests/org.eclipse.emfcloud.modelserver.client.tests/src/org/eclipse/emfcloud/modelserver/client/v2/ModelServerClientV2Test.java
```diff
@@ -0,0 +1,721 @@
...
+import com.google.common.base.Charsets;
+
+import okhttp3.HttpUrl;
+import okhttp3.HttpUrl.Builder;
+import okhttp3.OkHttpClient;
+import okhttp3.mock.MockInterceptor;
+import okhttp3.mock.Rule;
+import okio.Buffer;
+
+public class ModelServerClientV2Test {
```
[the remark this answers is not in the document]
CamilleLetavernier: All relevant classes have been forked when we started splitting the API. However, since we didn't actually provide a V2 Client for Java, the V2 Client and the V2 Tests are completely identical to V1 (Except for file names)

pom.xml
ndoschek: Just as a small sidetrack here: For us it is in general easier if we stick to the variable when we want to publish a release (candidate) as we can then use our scripts to adapt the versions without having to check manually.

CamilleLetavernier: If the target platform itself is not published, you can achieve the same result by using a fixed version (There's no need to ever change the target platform version; you can set it to `1.0.0` or `0.0.1` and forget about it).

Alternatively, you can use an explicit version Property (instead of the predefined `${project.version}`), to keep the Target Platform versioned. This way, the Target Platform can still inherit the version from the parent, but projects referencing the target platform may be versioned separately (while still using this shared target platform at a specific version).

The idea is really that the target platform should have its own version, rather than reuse the version from the project referencing it.

bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/client/AbstractModelServerClient.java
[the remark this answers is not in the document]
eneufeld: as we decided to do this separately, closing all comments about V1/V2 duplications

eneufeld (approved):
Looks good to me, I suggest to remove the draft flag.

examples/org.eclipse.emfcloud.modelserver.example/src/main/java/org/eclipse/emfcloud/modelserver/example/client/ExampleModelServerClient.java
```diff
@@ -53,7 +53,7 @@ private ExampleModelServerClient() {}
 
    public static void main(final String[] args) {
       try (
-         ModelServerClient client = new ModelServerClient("http://localhost:8081/api/v1/",
+         ModelServerClientV2 client = new ModelServerClientV2("http://localhost:8081/api/v2/",
```
ndoschek, line 56: Could you please have another look at this example client?

I have the following problems:
1. If subscription is closed due to server shutdown an Exception is thrown, previously the reason was logged.
2. The edit requests fail with `java.lang.RuntimeException: Could not parse 'data' field`
3. The undo/redo requests return `202: false Accepted`

ndoschek (changes requested):
Thanks again Camille and Christian for the updates! :tada: 

Looks good to me so far, however I still have a few minor comments:

1. Some of the `org.eclipse.emfcloud.modelserver.emf.tests` are failing and they also produce one one warning in the workspace. The `*.emf.tests` are not executed in the p2 build currently (please see #167) but in Eclipse and for the m2 build they should be fixed. Could you please have another look?
2. The README should be updated before we merge this independently form the API documentation. It would be fine by me to create a section with the updated V2 API and put the V1 in a collapsible section or sth like that.

> > Will this story include API documentation as well?
> 
> That definitely makes sense. However, I think it's still a bit early for that, as some APIs and use cases are not clearly defined yet (Especially URI handling, which Json Patches we support, how responses/notifications are sent, and how to handle multi-resources use cases). But we should probably start working on an initial version soon. 

That sounds reasonable - could you please create an follow-up issue for the documentation story?

CamilleLetavernier:
I've merged the Log4J changes from master into the branch to solve conflicts (That was a big batch of conflicts).

I've also fixed the custom 'counter' endpoint example, which was only registered for the v2 API, and not for v1 anymore. Now, it's available on both versions. This adresses a bug reported on the client side (The custom example menus use v1)

ndoschek (approved):
Thanks a lot Camille for the quick fixes and the rebasing!
Special thanks for the great README update! :1st_place_medal: 
To me it looks good and ready to merge. Still I will add @martin-fleck-at as a final reviewer.

bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/client/ModelServerClientApiV2.java
```diff
@@ -0,0 +1,111 @@
...
+package org.eclipse.emfcloud.modelserver.client;
+
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+import org.eclipse.emfcloud.modelserver.command.CCommand;
+
+import com.fasterxml.jackson.databind.node.ArrayNode;
+
+public interface ModelServerClientApiV2<A> {
```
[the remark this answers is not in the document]
ndoschek: Issue created by Camille https://github.com/eclipse-emfcloud/emfcloud-modelserver/issues/170

bundles/org.eclipse.emfcloud.modelserver.edit/src/org/eclipse/emfcloud/modelserver/edit/util/CommandUtil.java
```diff
@@ -102,4 +117,200 @@ public static EObject createProxy(final EClass eClass, final String uri) {
       ((InternalEObject) eObject).eSetProxyURI(URI.createURI(uri));
       return eObject;
    }
+
+   public static Optional<CCommandExecutionResult> aggregateExecutionResults(
```
martin-fleck-at, line 121: Could you please elaborate what the purpose of this method is? It does not seem to be used anywhere.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/codecs/JsonCodecV2.java
martin-fleck-at: Could this be tracked in a task somewhere and referenced here so we do not forget about it?

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/ModelServerRoutingV2.java
```diff
@@ -0,0 +1,99 @@
...
+
+   @Override
+   protected void apiEndpoints() {
+      super.apiEndpoints();
+
+      put(ModelServerPathsV2.MODEL_BASE_PATH, this::setModel); // Was PATCH in V1
+
+      patch(ModelServerPathsV2.MODEL_BASE_PATH, this::executeCommand); // Was PATCH /edit in V1
+
+      post(ModelServerPathsV2.TRANSACTION, this::createTransaction);
```
martin-fleck-at, line 61: This seems like a whole new feature. Is there any documentation when and how it is used? I also did not see any dedicated support for this in the ModelServerClient on the Javascript side.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/DefaultSessionController.java
```diff
@@ -126,6 +141,7 @@ public void modelUpdated(final String modeluri) {
    }
 
    @Override
+   @Deprecated
```
martin-fleck-at, line 144: Do we really want to deprecate this API? Isn't it still valid V1 API without the Json patch?

README.md
```diff
@@ -103,9 +103,15 @@ It will output details for each request/response similar to the following snippe
 
 ### HTTP Endpoints
 
-If the model server is up and running, you can access the model server API via `http://localhost:8081/api/v1/*`.
+The Model Server supports two versions of the API (v1 and v2). If the model server is up and running, you can access the model server v1 API via `http://localhost:8081/api/v1/*`,
+and the v2 API via `http://localhost:8081/api/v2/*`
```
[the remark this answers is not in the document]
eneufeld: The readme is showing the current state. And we have 2 versions at the moment.
As long as we have no V2 Java Client we need both APIs.
At a near future we will have only one API and thus deprecate V1 and remove it at some point.

bundles/org.eclipse.emfcloud.modelserver.common/src/org/eclipse/emfcloud/modelserver/common/patch/AbstractJsonPatchHelper.java
```diff
@@ -0,0 +1,716 @@
...
+         // return null.
+         return null;
+      }
+   }
+
+   protected Command testAction(final ResourceSet resourceSet, final JsonNode patchAction)
+      throws JsonPatchTestException {
+      boolean isValid = true;
+      // TODO ... implement actual test
+      // Should we throw an exception, or return an UnexecutableCommand?
```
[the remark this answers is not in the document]
CamilleLetavernier: I think Exception gives us more flexibility, to generate e.g. a more detailed client reply. "UnexecutableCommand" doesn't say anything about why it didn't work; an Exception can hold more valuable information (And we probably want to return a different result based on this exception: e.g. a specific error code/message, rather than "OK but nothing changed" or a generic "Can't execute. Try something else.")

bundles/org.eclipse.emfcloud.modelserver.common/src/org/eclipse/emfcloud/modelserver/common/patch/JsonPatchTestException.java
```diff
@@ -0,0 +1,25 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+
+/**
+ * An exception thrown when a Json Patch "test" operation fails.
+ */
+public class JsonPatchTestException extends Exception {
+
+   private static final long serialVersionUID = 8333555068740826391L;
+
+   public JsonPatchTestException(final JsonNode testOperation) {
```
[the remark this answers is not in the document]
CamilleLetavernier: This should be implemented properly if/when we support test operations. If the Exception is meant to give detailed error informations, it needs to know what was being tested to know what failed. But since it's not really implemented, the parameter is currently unused. I don't know if the detailed error message will come from the exception itself, or from the class that creates the exception. In one case, it makes sense to have the test exception; in the other case it can be replaced with a simple preformatted string parameter.

bundles/org.eclipse.emfcloud.modelserver.edit/src/org/eclipse/emfcloud/modelserver/edit/util/CommandUtil.java
CamilleLetavernier: It is used with Transactions. When a transaction is open, we can execute multiple commands. Each command will generate a single CommandExecutionResult; but notifications will only be sent to listeners when the transaction is closed. At this point, we need to aggregate all available CommandExecutionResults into one. If it's not used (yet), then maybe a downstream project needs it (It's a public API after all). Maybe @cdamus can give more accurate info about this.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/ModelServerRoutingV2.java
CamilleLetavernier: It's used by a new project: https://github.com/eclipse-emfcloud/modelserver-node

We decided to exclude it from the Javascript/Java client API at the moment; it's considered an implementation detail, to be used through the `modelserver-node` APIs (At the moment. But maybe it could be properly opened in the future).

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/DefaultSessionController.java
CamilleLetavernier: Yes, we want to deprecate it. This method would only notify V1 clients, even if V2 Clients are listening to the model changes. The new method is the correct implementation, that works with either V1 or V2, or both. V1 implementations should use the new method too (Maybe with an empty supplier for the JsonPatch, if for whatever reason, a V1 Server already migrates to this V2 version of the Model Server API, without actually implementing V2).

The idea is that the Server should support V1 **and** V2 as much as possible. Clients are free to use either V1 or V2 APIs, and they don't need to know if the Server is a V1 or a V2 server (Because a V2 Server stills supports V1, at least for now).

bundles/org.eclipse.emfcloud.modelserver.common/src/org/eclipse/emfcloud/modelserver/common/patch/AbstractJsonPatchHelper.java
```diff
@@ -0,0 +1,716 @@
...
+               continue;
+            }
+
+            Object emfValue = getEMFValue(modelURI, resourceSet, feature, field.getValue());
+            newObject.eSet(feature, emfValue);
+         }
+         return newObject;
+      }
+
+      // Other EClassifiers: Enum and Datatype can't be instantiated (But they may be used to specify values with
```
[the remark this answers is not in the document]
CamilleLetavernier: This method "creates" new elements. "creating" an enum literal doesn't make sense.

To set an enum value, we would use a "replace" operation rather than "add" (See `getEMFValue` and `getEMFEnumValue`). 

Regarding EDataTypes, there is a follow-up issue (Issue #162). It would make sense to use the createFromString() method in that context (But probably not in the `getObjectToAdd` operation; this would be a "replace" operation too. I think. There are subtle differences between "Json Patch AddOperation" and "EMF Add Command").

I guess it mostly depends on how we represent DataTypes on the json side. If they are represented a Opaque Strings, then we should use "replace". If they are specified as expanded types, then we should use "Add", and indeed add DataType support in the getObjectToAdd method (But then we wouldn't be able to use "createFromString", as we'd be receiving a structured object). Anyway, there are many special cases to consider to properly support EDataTypes, so having this discussion in the appropriate ticket (#162) makes more sense :)

README.md
```diff
@@ -103,9 +103,15 @@ It will output details for each request/response similar to the following snippe
...
 
-The following table shows the current HTTP endpoints:
+#### HTTP Endpoints - V1
+
+<details>
```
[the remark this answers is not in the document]
CamilleLetavernier: I don't have a strong preference. Just following @ndoschek 's suggestion. Although having two very similar tables in the same page would IMO be confusing, as you would always have to double-check you're looking at the correct version. Having a more detailed REST API Documentation would make sense (@ndoschek suggested using Javalin OpenAPI in #170), and then we could remove these tables entirely from the README (Or replace them with the generated pages, depending on how large they are; but I think referencing the pages makes more sense).

bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/client/TransactionContext.java
```diff
@@ -0,0 +1,41 @@
...
+
+   /**
+    * Roll back all commands executed so far in the context of the transaction and close it without putting a compound
+    * on the stack.
+    *
+    * @param error
+    */
+   void rollback(Diagnostic error);
+
+   // TODO: Call-back for receipt of the CommandExecutionResult
```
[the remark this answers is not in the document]
CamilleLetavernier: @cdamus I leave this one to you :)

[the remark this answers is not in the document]
cdamus: This part of the Client API is incomplete, only partially implementing support for execution of commands as indicated by this comment looking for a way to report execution results back to the client. There isn't even any mention of the JSON Patch method for API v2 here, which the endpoint implementation in the server supports.

As it is not currently the intention of this contribution to support the transaction endpoint for the general class of clients, I think I would actually recommend removing this interface entirely together with other mentions of the endpoint, in this PR:

- `TransactionContext` interface and its implementations
- `ModelServerClientApiV2::createTransaction(String)` method and its implementation
- protected APIs of the `ModelServerClientV2` class supporting transactions

bundles/org.eclipse.emfcloud.modelserver.edit/src/org/eclipse/emfcloud/modelserver/edit/util/CommandUtil.java
cdamus: It is indeed unused, almost certainly a vestige of some iteration of the code before this state. I'm happy to remove it.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/ModelServerRoutingV2.java
cdamus: Yes, it's intended to be an internal API used by a coöperating network of model servers only, not by client applications. Perhaps that's a concept that could be formalized in the _Client API_ if we want to include the transaction endpoint in it? But we do not foresee other such internal APIs and as mentioned it isn't certain that this will not be opened to all clients when we've a clear idea of how that should work.

cdamus:
Thanks for your review, Martin! You raise several good questions.

> I tested it as far as I could with the v1 Theia client and the v1 Java client but I could not find any good way to test the v2 API. From the code point of view, it looks good but I do have a few comments and one question: How do we plan to proceed with V1? Some code suggests more a dual approach where both APIs can be used independent from each other but other classes mix their implementation together. I thought we could split the V1 vs V2 based on the server modules that are created, i.e., that there will be a V2 DefaultSessionController etc. which are bound in a V2 DefaultModelServerModule so that adopters can either choose that server module as their base or the V1 one. For instance, the current approach leads to the DefaultSessionController having a method called `subscribeV2` and two maps that store the listeners based on the version. It just strikes me as a bit odd so I am curious, what was the reasoning behind this?

In the short term, it is expected that applications will continue to use the v1 API if it works for them. Of course, early adopters of v2 can help to shake things out there, but for now it may still undergo some churn as use cases are fleshed out and gaps in the functionality filled. In the longer term, the intent is that the v1 API can be deprecated and eventually phased out.

On the matter of how the API split is realized in the server, there is so much commonality in the shape and behaviour of these two API versions that it seemed most practical to share the bulk of the implementation and distinguish the divergent behaviour where necessary as in this case in the `DefaultSessionController`. The server needs to continue to provide the API v1 endpoints, so the proposal is not to make applications choose one version or the other by selecting which Guice module to use. If it is anticipated that adopters will need to jettison the v1 API from their applications by such a mechanism now, then perhaps we can have separate modules that bind the existing services as they are to their respective routes, and also a third module that combines the bindings from both, which would be the default module. What do you think?

cdamus:
Commit 97513051db5a37a356ba732976ce39718f7f24d5 adds changes to address specific review comments:

- remove unused code in the `CommandUtil` class
- remove incomplete and unsupported client API for transactions
- document the internal transaction endpoint in the readme

README.md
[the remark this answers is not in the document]
martin-fleck-at: Ok, thank you for clarifying that for me. Of course a generated API of some sort would be even better. I do not have too strong an opinion on this and since there is an open documentation task I am fine with how things are at the moment.

README.md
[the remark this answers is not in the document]
martin-fleck-at: Alright, let's hope that "in the near future" is not too far away. And since there is a proper documentation task #170 for v2 I think it is fine to leave the README like that for now. Still, I believe it might be a bit confusing for others why there are two versions of the API all of a sudden.

bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/client/TransactionContext.java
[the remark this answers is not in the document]
martin-fleck-at: I agree with the removal which already was added to this PR, thank you very much @cdamus!

bundles/org.eclipse.emfcloud.modelserver.common/src/org/eclipse/emfcloud/modelserver/common/patch/AbstractJsonPatchHelper.java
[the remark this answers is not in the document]
martin-fleck-at: I agree that in general an exception may have more information but this exception really does not ;-) Also currently we do not even through an exception. @CamilleLetavernier Could you maybe add a task to track adding this feature? As otherwise this TODO just seems like we were too lazy to do it instead of consciously postponing the implementation.

bundles/org.eclipse.emfcloud.modelserver.common/src/org/eclipse/emfcloud/modelserver/common/patch/AbstractJsonPatchHelper.java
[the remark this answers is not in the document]
martin-fleck-at: @CamilleLetavernier Thank you very much for explaining that to me. I was not aware of those fine differences between adding/creating and replacing in that context. Seeing as there is already a ticket for it, I am fine to have it like this for now since it does not prevent the rest from working as expected.

bundles/org.eclipse.emfcloud.modelserver.common/src/org/eclipse/emfcloud/modelserver/common/patch/JsonPatchTestException.java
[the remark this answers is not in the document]
martin-fleck-at: Understood, let's leave it for now.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/ModelServerRoutingV2.java
martin-fleck-at: I see, thank you for the explanation. I think removing the entries from the default paths to not explicitly expose it was the right thing to do for now. Thank you for that!

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/DefaultSessionController.java
martin-fleck-at: Hmm... I see though I might argue a bit differently that if the server should support both versions, adopters that already use that method should not necessarily get a deprecation warning since both calls actually still work but I might be missing something. Hopefully we can deprecate v1 in the near future altogether. Since the PR is already quite big and this is just a minor thing, I am fine with leaving it as is.

martin-fleck-at (approved):
@CamilleLetavernier @cdamus I had a final look at the changes and with clarifications you both gave and the additional "removal" of the transactions from the API, I am more than happy with this change. Thank you very much for your great work!

> In the short term, it is expected that applications will continue to use the v1 API if it works for them. Of course, early adopters of v2 can help to shake things out there, but for now it may still undergo some churn as use cases are fleshed out and gaps in the functionality filled. In the longer term, the intent is that the v1 API can be deprecated and eventually phased out.
> [...]

Thank you very much for the detailed reasoning behind this. In that sense, V2 is more of an extension of V1 than an actual standalone version which is fine for me. We can definitely leave the separation of versions in modules for another day - if ever - since V1 still works and will be deprecated sometime in the (near) future anyway. It just caught me by surprise since in my mental model it was more of an either/or situation.

CamilleLetavernier:
I've squashed the PR; the code is still exactly the same.

fixes #150

PR #158, opened 2022-02-02
#148 #6 Update to latest log4j2 version and improve Log Configuration

### LINKED ISSUE #6: Improve Log Configuration
We need to improve our Logging configuration at some point. Currently, the Logger just forwards to `System.Out`. Ideally, errors should be printed to `System.error` so I can react in the Theia Backend by listening to the error stream. However, for now, we bypassed this problem by setting the log level to error via CLI argument.

### LINKED ISSUE #144: Update Javalin dependency
Currently we are using version 4.0.0-ALPHA0. This effectively blocks any upcoming release(s). For release the Eclipse Foundation doesn't allow to include any unreleased content or snaphots builds of other projects (see: https://www.eclipse.org/projects/handbook/#ip-faq). Therefore we should update to the latest Javalin release (Currently 4.1.1). I quickly tested it and it looks like we need to do some minor refactoring because breaking changes we released between the 4.0.0-Alpha and the latest version.

### LINKED ISSUE #148: Update to latest log4j 2.x version
In light of the recent logj zero-day issues the Eclipse Foundation recommends to upgrade all log4j dependencies.
We are currently using 1.x version which means we are not directly affected by the current security leak. However, version 1.x is no longer maintained and the Eclipse Foundation recommends to upgrade anyways.

ndoschek:
#148 Update to org.apache.logging.log4j version to 2.17.1
- consumes from intermediate orbit build - desired version will be available with 2022-03 release (follow-up: #166)
- Migrate to log4j v2

#144 Upgrade to stable Javalin 4 version
- Upgrade to Javalin 4.3.0
   - Adapt to API changes
   - JavalinJackson is no Singleton anymore - therefore we consume the ObjectMapper via the existing ProviderDefaults class instead
 
#6 Improve Log Configuration
- Unify logging mechanisms and use log4j-slf4j-impl to redirect Javalin's slf4j logging to Log4j2 logger (as Log4J offers more flexible configuration)
- Introduce log4j2.xml configuration files
  - Default configuration on level 'debug' and prints fatal/error to stderr, all levels above to stdout
  - Add configuration files to classpath
- Introduce new CLI arguments to declare custom log configuration file and to enable extensive developer request/response logging
- Remove errorsOnly CLI argument
- Use complete classname for loggers - in default logging configuration it is configured to only use simple class name

General
- Update org.eclipse.emfcloud.emfjson-jackson version to 2.0.0
- Fix build warnings and correct output folders in build.properties
- Add ping option to ExampleModelServerClient
- Upgrade maven dependencies to match target platform
   - Update dependency jars in org.eclipse.emfcloud.modelserver.lib
- Update README (and resolve some markdownlint warnings)

Fixes #6
Fixes #144
Fixes #148


martin-fleck-at (commented):
@ndoschek Thank you very much for this change! I tested it and everything still looks good to me, so I believe we can merge this as is.

The only thing I noticed that is now even more obvious is the usage of the object mapper: We seemingly bind it in the `DefaultModelServerModule` using `ProviderDefaults.provideObjectMapper()` but then we never actually inject the mapper somewhere but always use the default object mapper instead. I just had a quick look and it seems that there are still a few classes the simply use static access, e.g., the `JsonCodec`. So maybe this is something to consider at some point, what do you think? This definitely does not have to be part of this change which already covers a lot. I just want to get your opinion on it before I create a ticket.

ndoschek:
Thanks for your review Martin!
Yes indeed, we never inject it at the moment, only use the static access in static methods of `JsonCodec` or the util `ContextRequest`.
Opening an issue to investigate that further would be a good idea imo. Thanks!

martin-fleck-at (approved):
Thank you Nina! Let's merge this and open another issue for the injection. Great work!

Issue #156, opened 2022-01-31
GetAll Method from the ModelServerClient.java does not return our models


NomsaT:
Currently trying to access all the models from our workspace using java API but the getAll method returns an empty list + some warnings in the console indicating that the models are incomplete. By calling the API directly from Postman, it retrieves all models without problem.
Does anyone know if our models need to be in a specific format for them to be "complete" because from our observations the two conditions for the if statement inside the getAll method are false( i.modelUri.isPresent() && content.isPresent()) and we are not sure why this is the case.

vhemery:
I made a rapid guess analysis on https://github.com/eclipse-emfcloud/emfcloud/discussions/113
To be confirmed...

martin-fleck-at:
@NomsaT Thank you very much for reporting this! This is indeed a wrong implementation in the Java ModelServerClient, I'll provide a fix for it soon. The Theia client uses the [correct parsing method](https://github.com/eclipse-emfcloud/emfcloud-modelserver-theia/blob/79b6e7b1acd705f42daebd5487de765bf4a15588/packages/modelserver-client/src/utils/type-util.ts#L141).

@vhemery Thank you for providing such a quick response and analysis! This really helped to narrow down the problem a lot.

ndoschek:
Closed via #164

PR #164, opened 2022-02-03
#156: Properly parse models from 'getAll' in Java ModelServerClient

martin-fleck-at:


examples/org.eclipse.emfcloud.modelserver.example/src/main/java/org/eclipse/emfcloud/modelserver/example/client/ExampleModelServerClient.java
```diff
@@ -149,6 +155,24 @@ private static void handleGet(final ModelServerClient client, final String[] com
...
+         System.out.println(response.body().size() + " models");
+         for (Model<String> model : response.body()) {
+            System.out.println("- " + model.getModelUri() + ": " + model.getContent());
+         }
+      } else {
+         Response<List<Model<EObject>>> response = client.getAll(format).join();
+         System.out.println(response.body().size() + " models");
+         for (Model<EObject> model : response.body()) {
+            System.out.println("- " + model.getModelUri() + ": " + model.getContent().eClass());
+         }
```
ndoschek, line 172: Why do we need the distinction here? If the `xmi` model content is the same string output as the `eClass` output in my case.

ndoschek (changes requested):
Thanks a lot for the fix Martin! Looks good to me!
I do have a minor comment (please see inline) and one small request: Could you please update the year ranges in the file headers? Thanks!

examples/org.eclipse.emfcloud.modelserver.example/src/main/java/org/eclipse/emfcloud/modelserver/example/client/ExampleModelServerClient.java
martin-fleck-at: I'm not sure what you mean by the same content but I split this method to showcase the difference between the json case where you get "only" the string and the 'xmi' where you get the actual EObject and thus can access the eClass.

ndoschek: Sorry for the confusion, I overlooked the format argument difference here.

PR #169, opened 2022-02-14
API v2: Prefer delegation over inheritance

### LINKED ISSUE #168: API v2: Share implementations by delegation
If we pull out an abstract client which is extended by V1 and V2 then we should pull out parameters. Otherwise this is weird imho.

_Originally posted by @eneufeld in https://github.com/eclipse-emfcloud/emfcloud-modelserver/pull/151#discussion_r800438406_

----

The API v1 and v2 implementations of the following components currently share common behaviour by inheriting common base classes. It is preferable to share these by delegation to avoid awkwardness of deprecation/de-implementation and incompatible API signatures as the v2 API evolves and in the future other versions (2.1, 3.0, ...?) are added.

- ModelServerClient interface and implementation classes
- ModelServerRouting implementation classes

cdamus:
Split implementations of v1 and v2 APIs via delegation to a common partial implementation instead of inheritance of a common partial implementation in

- the `ModelServerClient` API
- the `ModelServerRouting` handlers

The previously existing abstract base classes are deleted (actually, more or less renamed and moved) instead of deprecated because they were only actually introduced during this release in PR #151, so there is no concern about breaking API for downstream applications.

Fixes #168.

### DIFF
--- a/bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/client/ModelServerClient.java
+++ b/bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/client/ModelServerClient.java
@@ -20,7 +20,7 @@ import okhttp3.OkHttpClient;
 
 public class ModelServerClient extends ModelServerClientV2 {
 
-   public static final Set<String> DEFAULT_SUPPORTED_FORMATS = ModelServerClientV2.DEFAULT_SUPPORTED_FORMATS;
+   public static final Set<String> DEFAULT_SUPPORTED_FORMATS = ModelServerClientV2.SUPPORTED_FORMATS;
    public static final String PATCH = ModelServerClientV2.PATCH;
    public static final String POST = ModelServerClientV2.POST;
 
--- a/bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/client/v1/ModelServerClientV1.java
+++ b/bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/client/v1/ModelServerClientV1.java
@@ -11,88 +11,285 @@
 package org.eclipse.emfcloud.modelserver.client.v1;
 
 import java.net.MalformedURLException;
+import java.util.List;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.eclipse.emf.ecore.EObject;
-import org.eclipse.emfcloud.modelserver.client.AbstractModelServerClient;
 import org.eclipse.emfcloud.modelserver.client.EditingContext;
+import org.eclipse.emfcloud.modelserver.client.Model;
 import org.eclipse.emfcloud.modelserver.client.ModelServerClientApiV1;
 import org.eclipse.emfcloud.modelserver.client.Response;
+import org.eclipse.emfcloud.modelserver.client.ServerConfiguration;
+import org.eclipse.emfcloud.modelserver.client.SubscriptionListener;
 import org.eclipse.emfcloud.modelserver.command.CCommand;
 import org.eclipse.emfcloud.modelserver.common.ModelServerPathParametersV1;
 import org.eclipse.emfcloud.modelserver.common.ModelServerPathsV1;
 import org.eclipse.emfcloud.modelserver.emf.common.JsonResponseMember;
 import org.eclipse.emfcloud.modelserver.emf.configuration.EPackageConfiguration;
-import org.eclipse.emfcloud.modelserver.internal.client.EditingContextImpl;
+import org.eclipse.emfcloud.modelserver.internal.client.ModelServerClientDelegate;
 import org.eclipse.emfcloud.modelserver.jsonschema.Json;
 
 import okhttp3.MediaType;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.RequestBody;
-import okhttp3.WebSocket;
 
-public class ModelServerClientV1 extends AbstractModelServerClient
-   implements ModelServerClientApiV1<EObject>, ModelServerPathsV1 {
+public class ModelServerClientV1 implements ModelServerClientApiV1<EObject>, ModelServerPathsV1, AutoCloseable {
 
-   public static final Set<String> DEFAULT_SUPPORTED_FORMATS = AbstractModelServerClient.DEFAULT_SUPPORTED_FORMATS;
+   public static final Set<String> SUPPORTED_FORMATS = Set.of(ModelServerPathParametersV1.FORMAT_JSON,
+      ModelServerPathParametersV1.FORMAT_XMI);
 
    protected static final Logger LOG = LogManager.getLogger(ModelServerClientV1.class);
+
+   private final ModelServerClientDelegate delegate;
 
    public ModelServerClientV1(final String baseUrl, final EPackageConfiguration... configurations)
       throws MalformedURLException {
 
-      super(baseUrl, ModelServerPathParametersV1.FORMAT_JSON, configurations);
+      this(new ModelServerClientDelegate(baseUrl, ModelServerPathParametersV1.FORMAT_JSON, SUPPORTED_FORMATS,
+         configurations));
    }
 
    public ModelServerClientV1(final OkHttpClient client, final String baseUrl,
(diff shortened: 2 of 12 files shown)


ndoschek (commented):
Thanks a lot @cdamus ! I had a look at your [commit](https://github.com/eclipse-emfcloud/emfcloud-modelserver/commit/135df52b823a94caea1442f572af5cd3000c4641) and it looks great to me!
I will re-test it once we merged the base PR #151 .

cdamus:
Commit 8181a0718083f0571006832c60a9d4e715262e57 rebases the branch onto the latest log4j refactoring in branch `/issues/150-apiV2`.

cdamus:
Commit 1c3b2d775ac2c372656d0653e4bcb6e9315c05d6 rebases this PR onto the latest changes from #151 which now has essentially finished its review.

ndoschek (changes requested):
Hi again @cdamus,
Thanks for resolving conflicts an dupdating the PR!
I had a look and changes look good to me!
I spotted an issue with the `ExampleModelServerClient` (see inline) - could you please have another look?

And I have a question regarding the default format of requests:
V1: If I send an `edit` request (containing an EMF command in JSON for example) without a dedicated `format` parameter, an Exception is thrown, same for V2.
I think I can remember something about default formats but I can't find the conversation about that anymore. Could you please refresh my memory on that? Thanks :)

cdamus:
> Hi again @cdamus, Thanks for resolving conflicts an dupdating the PR! I had a look and changes look good to me! I spotted an issue with the `ExampleModelServerClient` (see inline) - could you please have another look?

Thanks for picking this up, @ndoschek .

The problem is threefold:

1. The example was updated to use `json-v2` format with the V1 client implementation. That just doesn't make sense because `json-v2` format is intended for the V2 API.
2. The client API (both v1 and v2 implementations) encode the command in the `"data"` property of the `edit` request body within a JSON string, which the server doesn't expect when it decodes the request.
3. The `JsonCodecV2` overrides the `encode(EObject)` method to skip the copy into a one-off `JsonResource` because that caused problems with serialization of `$id` properties. But this breaks the encoding of an edit request in a `CCommand` has a proxy reference to the `owner` because without the containing resource context, the object mapper won't look into the proxy URI to get the URI fragment and so we end up with a `"$ref": null` that bombs the command in the server end.
 
So, I've added a commit that fixes all of these problems:

- using the correct client API in the example
- checking in the server end for a string-valued `"data"` in the `edit` message and parsing that as a JSON object
- using the default method of copying into a temporary resource to encode an object that isn't in a resource to begin with and so cannot possibly have useful IDs

@CamilleLetavernier this last point makes the execution of commands via the Java client API work, but I'm not sure that it doesn't break some other use case by introducing `"$id": null` into the JSON. Can you check this with whatever use case you originally had to work around in the `JsonCodecV2`? TIA.

Anyways, with this new commit I can have an interaction with the server in the example client like the snippet quoted at at the bottom of this comment.

> And I have a question regarding the default format of requests: V1: If I send an `edit` request (containing an EMF command in JSON for example) without a dedicated `format` parameter, an Exception is thrown, same for V2. I think I can remember something about default formats but I can't find the conversation about that anymore. Could you please refresh my memory on that? Thanks :)

Unless I'm mistaken, this concept of "default format" is implemented in the server and in the Javascript client API. I don't think anything like it is implemented in the Java client API here. Perhaps it should be, but that would be a separate enhancement request.

----

```console

> subscribe SuperBrewer3000.json
< OK
2022-03-02T20:00:06.658Z | [SuperBrewer3000.json]: Connected: Switching Protocols
2022-03-02T20:00:06.680Z | [SuperBrewer3000.json]: Success: df8e42f7-8818-4350-926d-203d5e240d67
2022-03-02T20:00:06.691Z | [SuperBrewer3000.json]: Dirty State: false

> update-tasks Damus
2022-03-02T20:00:17.444Z | [SuperBrewer3000.json]: Incremental <JsonString> update from model server received: [{"op":"replace","path":"/workflows/0/nodes/0/name","value":"PreHeat Damus"}]
2022-03-02T20:00:17.444Z | [SuperBrewer3000.json]: Dirty State: true
< 200: true OK

> get SuperBrewer3000.json
< SuperBrewer3000.json
{
  "$type" : "http://www.eclipsesource.com/modelserver/example/coffeemodel#//Machine",
  "$id" : null,
  "children" : [ {
    "$type" : "http://www.eclipsesource.com/modelserver/example/coffeemodel#//BrewingUnit",
    "$id" : null
  }, {
    "$type" : "http://www.eclipsesource.com/modelserver/example/coffeemodel#//ControlUnit",
    "$id" : null,
    "processor" : {
      "$id" : null,
      "clockSpeed" : 5,
      "numberOfCores" : 10,
      "socketconnectorType" : "Z51",
      "thermalDesignPower" : 100
    },
    "display" : {
      "$id" : null,
      "width" : 10,
      "height" : 20
    }
  } ],
  "name" : "Super Brewer 3000",
  "workflows" : [ {
    "$id" : null,
    "name" : "Simple Workflow",
    "nodes" : [ {
      "$type" : "http://www.eclipsesource.com/modelserver/example/coffeemodel#//AutomaticTask",
      "$id" : null,
      "name" : "PreHeat Damus",
      "component" : {
        "$type" : "http://www.eclipsesource.com/modelserver/example/coffeemodel#//BrewingUnit",
        "$ref" : "//@children.0"
      }
    } ]
  } ]
}

> 
```

CamilleLetavernier:
> @CamilleLetavernier this last point makes the execution of commands via the Java client API work, but I'm not sure that it doesn't break some other use case by introducing "$id": null into the JSON. Can you check this with whatever use case you originally had to work around in the JsonCodecV2? TIA.

For me, it was originally the opposite issue: having an object in a resource resulted in $id: null, because moving to a separate resource broke IDs.

Reusing the original resource when there is one, and a new resource when the object is detached, makes sense to me. I've tested a collection of Postman requests against your branch, and as far as I can tell, everything works as expected.

`$id: null` shouldn't be an issue in most cases (Especially for serialized objects that are not actual model elements, such as Commands). We expect a valid $id when creating Operations for a Model Element (Unless we use true json patches, in which cases $id are not used). We also test for the presence of an $id attribute in the TS Client (So in general, the server should try to map IDs to all objects, even if they're just positional IDs).

```
> get SuperBrewer3000.json
< SuperBrewer3000.json
{
  "$type" : "http://www.eclipsesource.com/modelserver/example/coffeemodel#//Machine",
  "$id" : null,
   ...
}
```

In this case, since it's an actual model, I would expect the Machine to have a non-null ID. Even a dynamic one such as "/". Is that because the object doesn't belong to a resource? But how was it loaded, then?

cdamus:
Thanks, Camille. That is reassuring.

> ```
> > get SuperBrewer3000.json
> < SuperBrewer3000.json
> {
>   "$type" : "http://www.eclipsesource.com/modelserver/example/coffeemodel#//Machine",
>   "$id" : null,
>    ...
> }
> ```
> 
> In this case, since it's an actual model, I would expect the Machine to have a non-null ID. Even a dynamic one such as "/". Is that because the object doesn't belong to a resource? But how was it loaded, then?

Yeah, this is odd. I need to look into it. All of the objects in this snippet have null `$id`s, but cross-references (there are not many) look to have correct `$ref`s.

cdamus:
> Yeah, this is odd. I need to look into it. All of the objects in this snippet have null `$id`s, but cross-references (there are not many) look to have correct `$ref`s.

OK, this was easy to answer. The secret is that in the snippet I quoted, I used the example client to get the model from the `*.json` resource, not the `*.coffee` resource, which latter is an XMI resource. I used the JSON resource because that's the one that the example command targets.

When I get the model from the XMI resource, it works as expected:

``` console
> get SuperBrewer3000.coffee
< SuperBrewer3000.coffee
{
  "$type" : "http://www.eclipsesource.com/modelserver/example/coffeemodel#//Machine",
  "$id" : "/",
  "children" : [ {
    "$type" : "http://www.eclipsesource.com/modelserver/example/coffeemodel#//BrewingUnit",
    "$id" : "//@children.0"
  }, {
```

So, there is no problem here and I'm satisfied with this PR, thanks to Camille's earlier reassurance on the matter of the `JsonCodecV2`.

ndoschek (approved):
Thanks Christian for the update and clarifications!
To me it looks great and works as I would expect it to :tada: 

> Unless I'm mistaken, this concept of "default format" is implemented in the server and in the Javascript client API. I don't think anything like it is implemented in the Java client API here. Perhaps it should be, but that would be a separate enhancement request.

Yes I agree, for v2 it is probably a separate discussion. I double checked and for v1 I think it is a regression we overlooked during the last PR #151. Therefore I created a separate bug: #182

PR #175, opened 2022-02-25
Fix #174 API v2: Injection of JsonPatchHelper

### LINKED ISSUE #174: API v2: Injection of JsonPatchHelper
In my application's implementation of the _Model Server_, I have a customization of the EMF <--> JSON transformation that results in JSON with a tree structure quite different to the EMF tree. Accordingly, I need to customize the both the transformation of incoming JSON patches for edit requests and outgoing diffs in the `ModelUpdateResult`s from edits.

Currently, a few components of the _Model Server_ construct a `JsonPatchHelper` for themselves to accomplish these tasks. I need to be able to inject my own patch helper to handle my custom EMF/JSON transformation.

This issue is of course only relevant in the context of API v2, which is pending in issue #150.

cdamus:
Inject the `JsonPatchHelper` also into the resource manager and transaction controller that previously constructed it for themselves.

This requires additionally:

- provider injection for the resource manager because the patch helper depends on the resource manager
- configuration of injectors in tests now needs to include injection of codecs map because it is injected into the patch helper

So far, this is tied to `json-v2` format exclusively as this is the default format for API v2. If the helper should need to pick up other formats from the request/socket context, then that's an orthogonal issue.

### DIFF
--- a/bundles/org.eclipse.emfcloud.modelserver.common/src/org/eclipse/emfcloud/modelserver/common/codecs/Codec.java
+++ b/bundles/org.eclipse.emfcloud.modelserver.common/src/org/eclipse/emfcloud/modelserver/common/codecs/Codec.java
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2019 EclipseSource and others.
+ * Copyright (c) 2019-2022 EclipseSource and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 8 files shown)


cdamus:
Commit d2c041d5a812e054846567b6823a06ab2f134fbf rebases the PR onto the latest state of the #151 branch, which has essentially completed its review.

cdamus:
Commit d97e467422132c9400c779684d479dd5af02ab6b rebases the branch onto the latest from the `master` branch to resolve a merge conflict.

That conflict was in the `JsonPatchHelper::getCurrentModel(...)` method which on the `master` branch was changed to directly serialize the model using the object mapper from the `JsonCodecV2` class. However, this class is substitutable by Guice injection with a codec that uses a different mapper, which the current branch originally accounted for via the `CodecsManager` registry.  So, in resolving the conflict, I've introduced a new internal interface for codecs that does as direct as possible a serialization using that codec's mapper which we can expect to be available on all codecs that extend the `DefaultJsonCodec` (which in practice should be all of them).

cdamus:
> Should we maybe bind the JSONPatchHelper as part of the default bindings? Or is it already the case and I just missed it?

Guice does not permit self-bindings. As the `JsonPatchHelper` is a concrete class, it is creatable by default, so it is implicitly bound to itself by Guice. We could make it explicit with an incomplete binding like

```java
   bind(JsonPatchHelper.class);
```

but I think there are other cases in this project where we decided not to do that (?).

eneufeld:
> > Should we maybe bind the JSONPatchHelper as part of the default bindings? Or is it already the case and I just missed it?
> 
> Guice does not permit self-bindings. As the `JsonPatchHelper` is a concrete class, it is creatable by default, so it is implicitly bound to itself by Guice. We could make it explicit with an incomplete binding like
> 
> ```java
>    bind(JsonPatchHelper.class);
> ```
> 
> but I think there are other cases in this project where we decided not to do that (?).

So if it is autobound I'm happy.

cdamus:
Commit 9f38ae49488acd34aecce096beafff9a6988f5dd is rebased on the latest updates from the `master` branch for compatible extensibility refactorings.

ndoschek (approved):
Thanks a lot @cdamus, changes look great to me! :+1:

eneufeld (approved):
Looks good, thank you!

PR #185, opened 2022-03-08
#184 Improve extensibility

### LINKED ISSUE #184: Improve extensibility after introduction of v2
After the introduction of `v2` members and contexts are not extensible anymore, e.g.
- The field ModelServerClientV1.delegate is not visible
- The type ModelServerRoutingDelegate is not visible

ndoschek:
Based on the project `uml-glsp` some extensibility issues with V1 occured while updating to the latest version.

Resolves #184

### DIFF
--- a/bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/client/v1/ModelServerClientV1.java
+++ b/bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/client/v1/ModelServerClientV1.java
@@ -44,12 +44,20 @@ public class ModelServerClientV1 implements ModelServerClientApiV1<EObject>, Mod
 
    protected static final Logger LOG = LogManager.getLogger(ModelServerClientV1.class);
 
-   private final ModelServerClientDelegate delegate;
+   protected final ModelServerClientDelegate delegate;
 
    public ModelServerClientV1(final String baseUrl, final EPackageConfiguration... configurations)
       throws MalformedURLException {
 
       this(new ModelServerClientDelegate(baseUrl, ModelServerPathParametersV1.FORMAT_JSON, SUPPORTED_FORMATS,
+         configurations));
+   }
+
+   public ModelServerClientV1(final String baseUrl, final Set<String> supportedFormats,
+      final EPackageConfiguration... configurations)
+      throws MalformedURLException {
+
+      this(new ModelServerClientDelegate(baseUrl, ModelServerPathParametersV1.FORMAT_JSON, supportedFormats,
          configurations));
    }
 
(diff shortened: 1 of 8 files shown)


bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/client/v1/ModelServerClientV1.java
```diff
@@ -44,7 +44,7 @@
 
    protected static final Logger LOG = LogManager.getLogger(ModelServerClientV1.class);
 
-   private final ModelServerClientDelegate delegate;
+   protected final ModelServerClientDelegate delegate;
```
cdamus, line 47: This (together with making the delegate class public which would be required) is the only change that gives me pause. The idea of the delegate was that it was intended only to provide common behaviour of the core routing classes. Subclasses outside of the model server framework should not need to access the delegate directly, but only via APIs (preferably protected) of the V1/V2 classes. If necessary, those APIs should be added to the V1/V2 classes.

What specifically motivates opening this up to the world at large?

cdamus (approved):
Thanks @ndoschek for filling these gaps. In principal these changes look good to me, but I'm concerned about the need to expose something that I really intended should be internal to the core server framework, as indicated in the in-line comment. I'd like to understand better what that problem is and whether maybe there's a different solution that can continue to hide the delegate. But not at the cost of a lot of complication.

bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/client/v1/ModelServerClientV1.java
ndoschek: Thanks for your feedback on that @cdamus !
Oh I understand, yes this sounds of course very reasonable!

My specific use case is an extension of the `ModelServerClientV1` atm, which adds a custom request. It would therefore be great to reuse the `createHttpUrlBuilder` oder the `makeCallAndGetDataBody` methods (which are also restricted atm).
My code would look currently like this:

```
   @SuppressWarnings("restriction")
   public CompletableFuture<Response<List<String>>> getTypes(final String modelUri) {
      final Request request = new Request.Builder()
         .url(
            // createHttpUrlBuilder(delegate.getBaseUrl() + MyPaths.TYPES)
            delegate.createHttpUrlBuilder(baseUrl + MyPaths.TYPES)
               .addQueryParameter(ModelServerPathParametersV1.MODEL_URI, modelUri)
               .build())
         .build();

      return this.delegate.makeCallAndGetDataBody(request)
         .thenApply(response -> response.mapBody(body -> {
            List<String> uris = new ArrayList<>();
            try {
               Json.parse(body).forEach(uri -> uris.add(uri.textValue()));
               return uris;
            } catch (IOException e) {
               throw new CompletionException(e);
            }
         }));
   }
```
Should I extract this as a util method perhaps?

PR #186, opened 2022-03-08
#184 Improve backward API compability

cdamus:
- include the codecs that implement the supported formats in the client API
- add protected delegation APIs in public client and routing classes for continuity of access to APIs that existing clients rely on


ndoschek (approved):
Thanks a lot for the quick fix Christian! Works like a charm! :sparkle:

PR #189, opened 2022-03-14
Fix #188: URI Converter problems

### LINKED ISSUE #188: API v2: Problems with URI converter
The new `ModelURIConverter` service introduced in API v2 has two problems:

- a `GET` request on the `/api/v2/models` endpoint with a `modeluri` query parameter that is provided in the URL but that is rejected by the URI validator will return a mapping of all models by URI instead of the expected 404
- the URI converter was intended to ensure consistent handling of URIs but it forgot to account for a point of variability in `file` scheme URIs. In particular, URIs of the form `file:/path/to/resource.json` should be considered identical to URIs of the form `file:///path/to/resource.json` because an empty authority in a file URI is meaningless and some applications use an empty authority and some don't when creating file URIs from filesystem paths

The first of these problems is caused by the handler of the `GET` request delegating to the `ModelController` "get all models" API when the `modeluri` query parameter is omitted. The reason why this delegation happens in the case of a `modeluri` parameter that is specified but that is invalid is that the handler doesn't see the difference between an absent parameter and one that was ignored.

cdamus:
- delete empty authorities from incoming `file`-scheme model URIs
- double-check existence of the `modeluri` query param when delegating a `GET /models` request to `ModelController::getAll(Context)`


ndoschek (approved):
Thanks a lot Christian, looks great to me! :+1:

PR #194, opened 2022-03-18
193: JsonPatchHelper should normalize URIs when using EMF-style paths

### LINKED ISSUE #193: API V2: JsonPatchHelper should normalize URIs when using EMF-style paths
When using EMF-style paths with Json Patch, we include the resource URI in the path. Unlike the ?modeluri query, this URI is not normalized, which may cause some issues.

We should add URI normalization to JsonPatchHelper

CamilleLetavernier:
- Normalize URIs in ~DefaultModelResourceManager~ and JsonPatchHelper

fixes #193

### DIFF
--- a/tests/org.eclipse.emfcloud.modelserver.emf.tests/src/org/eclipse/emfcloud/modelserver/emf/common/watchers/FileModelWatcherIntegrationTest.java
+++ b/tests/org.eclipse.emfcloud.modelserver.emf.tests/src/org/eclipse/emfcloud/modelserver/emf/common/watchers/FileModelWatcherIntegrationTest.java
@@ -50,6 +50,7 @@ import org.eclipse.emfcloud.modelserver.emf.common.DefaultModelURIConverter;
 import org.eclipse.emfcloud.modelserver.emf.common.DefaultResourceSetFactory;
 import org.eclipse.emfcloud.modelserver.emf.common.ModelRepository;
 import org.eclipse.emfcloud.modelserver.emf.common.ModelResourceManager;
+import org.eclipse.emfcloud.modelserver.emf.common.ModelURIConverter;
 import org.eclipse.emfcloud.modelserver.emf.common.ResourceSetFactory;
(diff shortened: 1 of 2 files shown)


tests/org.eclipse.emfcloud.modelserver.emf.tests/src/org/eclipse/emfcloud/modelserver/emf/common/watchers/FileModelWatcherIntegrationTest.java
```diff
@@ -151,8 +152,8 @@ protected void configure() {
             bind(ModelRepository.class).to(DefaultModelRepository.class).in(Singleton.class);
             bind(ModelResourceManager.class).to(DefaultModelResourceManager.class).in(Singleton.class);
             bind(ResourceSetFactory.class).to(DefaultResourceSetFactory.class).in(Singleton.class);
-            bind(URIConverter.class).to(DefaultModelURIConverter.class).in(Singleton.class);
-
+            bind(ModelURIConverter.class).to(DefaultModelURIConverter.class).in(Singleton.class);
```
cdamus, line 155: Thanks for that!

cdamus (approved):
LGTM!

CamilleLetavernier:
I've reverted the changes to DefaultModelResourceManager, because it seems that all URIs should be normalized before they get passed to this class. Adding the ModelURIConverter to the DefaultModleResourceManager seemed to add unnecessary complexity (Especially for test setup), and was mostly redundant.

cdamus:
> I've reverted the changes to DefaultModelResourceManager, because it seems that all URIs should be normalized before they get passed to this class. Adding the ModelURIConverter to the DefaultModleResourceManager seemed to add unnecessary complexity (Especially for test setup), and was mostly redundant.

I think I reviewed this after you made that change? Because I didn't notice it and I would have commented that yes, it was unnecessary. 😉

sgraband (approved):
Thanks @CamilleLetavernier! LGTM :+1:

PR #196, opened 2022-04-04
API V2: Add Java implementation for the ModelServerClient

### LINKED ISSUE #172: API V2: Add Java implementation for the ModelServerClient
This is a follow-up for #151

The V2 API focuses on supporting web-native uses cases, by removing "EMFisms". The goal was to hide EMF as much as possible for Web clients, and support a self-sufficient Json Model, that can be edited with Json Patch.

For Java clients, however, using EMF is still perfectly fine, so the Java Client hasn't been updated for the V2 API (Although the V2 Server can support both use cases; EMF-based and Json-based).

At the moment, the following combinations are supported:

- V1 Java Client for V1 or V2 Java Server (The server implementation still supports both V1 and V2 in parallel)
- V1 TS Client for V1 or V2 Java Server
- V2 TS Client for V2 Java Server

We need to add proper support for a V2 Java Client, as V2 provides some improvements even for Java (such as Transactions). On the long run, it will also make it easier to slowly deprecate the V1 API and clean-up the implementations.

The current PR contains a stub for the Java Client V2 implementation, but it is essentially a fork of the V1 Client, and hasn't been fully updated to work with V2. It is also largely untested.

cdamus:
Add/refine the Java client for API v2 endpoints that correspond to API v1, but of course with their additional capabilities, especially JSON Patch for editing. This specifically excludes a client API for the `api/v2/transaction` endpoint which is intended for use only by the [modelserver-node layer][msnode].

- add an EMF implementation of the JSON Patch model to use in the client API
- add a codec-based subscription listener for payloads as `EObject`s, which enables Java-friendly access to the JSON Patch model update notifications
- update v2 client API signatures for new server capabilities, especially JSON patches in edit responses and subscriptions
- update example client to demonstrate new API v2 capabilities

Fixes #172.

[msnode]: /eclipse-emfcloud/modelserver-node/


cdamus:
The builds are now all broken by the removal of the Orbit repository.

ndoschek:
Hi @cdamus!
With #199 I updated the targetplatform to use the current Eclipse release `2022-03`. So at least the Jenkins Build should succeed again. With the GitHub action builds there are issues we should investigate (see also #200).
But if the local and Jenkins builds succeed we are good to go for now I'd say (although too nice of course). I will have a look at your PR today :eyes:

cdamus:
Hi, @ndoschek. Commit 47bbbdc120f0b3230fadfc52381c1bdb4842008e makes a bunch of changes to streamline the API and example application, pursuant to off-line discussion:

- rename the `CodecSubscriptionListener` class as `EObjectSubscriptionListener`
- rework inference of value types in JSON Patches  via a back-channel push of the model root `EClass` from the client
- refactor existing `AbcToEObjectSubscriptionListener` classes as subclasses of `EObjectSubscriptionListener`
- only parse incremental update as a JsonPatch in API v2 or later
  - this motivates some clean-up of `APIVersion` parsing and convenience constants
  - add specific subclass hooks for `JsonPatch` and `CCommandExecutionResult` variants of the incremental update
- rework the `ecore` pseudo-format for subscriptions in the example client as an optional "as eobject" command argument
  - support the same modifier in the `get `command
  - this also made the new shortcut for JSON-to-EObject subscription superfluous, so that is removed
- remove from the V2 client API the `edit(...)` method with `EditContext` web-socket that was never implemented on the server end
- update the README for v2 API changes:
  - the new V2 client interface summary
  - simple example of client API usage
  - example of subscription listener API usage

bundles/org.eclipse.emfcloud.modelserver.edit/resources/JsonPatch.genmodel
```diff
@@ -0,0 +1,62 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<genmodel:GenModel xmi:version="2.0" xmlns:xmi="http://www.omg.org/XMI" xmlns:ecore="http://www.eclipse.org/emf/2002/Ecore"
+    xmlns:genmodel="http://www.eclipse.org/emf/2002/GenModel" copyrightText="Copyright (c) 2022 STMicroelectronics and others.\n\nThis program and the accompanying materials are made available under the\nterms of the Eclipse Public License v. 2.0 which is available at\nhttps://www.eclipse.org/legal/epl-2.0, or the MIT License which is\navailable at https://opensource.org/licenses/MIT.\n\nSPDX-License-Identifier: EPL-2.0 OR MIT"
```
ndoschek, line 3: I guess this line is the cause for the checkstyle issues. In the `Command.genmodel` we use `&#xA;` instead of `\n`.
```suggestion
    xmlns:genmodel="http://www.eclipse.org/emf/2002/GenModel" copyrightText="Copyright (c) 2022 STMicroelectronics and others.&#xA;&#xA;This program and the accompanying materials are made available under the&#xA;terms of the Eclipse Public License v. 2.0 which is available at&#xA;https://www.eclipse.org/legal/epl-2.0, or the MIT License which is&#xA;available at https://opensource.org/licenses/MIT.&#xA;&#xA;SPDX-License-Identifier: EPL-2.0 OR MIT"
```

bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/client/TypedSubscriptionListener.java
```diff
@@ -20,9 +21,19 @@
...
-   private final Function<? super String, Optional<T>> updateFunction;
+   private final BiFunction<? super String, ? super String, Optional<? extends T>> updateFunction;
 
-   public TypedSubscriptionListener(final Function<String, Optional<T>> updateFunction) {
+   public TypedSubscriptionListener(final Function<String, Optional<? extends T>> updateFunction) {
+      this(updateFunction, Function.class);
+   }
+
+   public TypedSubscriptionListener(final Function<String, Optional<? extends T>> updateFunction,
+      final Class<?> discriminator) {
```
ndoschek, line 31: Do we need this `discriminator` here? I can't find a usage yet

ndoschek (changes requested):
Thanks a lot @cdamus, code looks great to me and works as expected! :tada:
Special thanks for updating the README and adding a batch of new tests! :+1: 
Regarding the checkstyle issues, I think I spotted the problem and added a comment inline.

bundles/org.eclipse.emfcloud.modelserver.edit/resources/JsonPatch.genmodel
cdamus: Oh, good find. Yes, it should be XML entities used here.

bundles/org.eclipse.emfcloud.modelserver.client/src/org/eclipse/emfcloud/modelserver/client/TypedSubscriptionListener.java
cdamus: Hah! Thanks for pointing this out. It Turns Out™ that the `discriminator` is now superfluous.

At one stage, I had two different constructors accepting a `Function` with different type arguments, which erase to the same signature and so needed an additional argument in one of them to distinguish the constructor signatures. But some later refactoring eliminated the second constructor an so now this dummy `discriminator` argument is not needed. Consider it deleted.

cdamus:
Commit 3b56b78c5f0241b43c93be356d4e5fb61bb3d3ba fixes the copyright headers in the generated code (Eclipse code folding FTW!) and removes the superfluous constructor argument in the `TypedSubscriptionListener` class. While I was in there, I thought I should add some doc comments.

Regenerating the model also fixed an inconsistency in the generated code's metadata for the `stringValue()` operation that originally had been named `valueString()` but which at some point I had renamed as `stringValue()`. But in that refactoring exercise I had neglected to regenerate the package metadata. So, happily that is now fixed.

cdamus:
@ndoschek there are still two Checkstyle problems in the EMF-generated doc comments in `XyzImpl` classes. It is strange because

- these doc comments are functionally empty like so many others in an EMF-generated implementation class
- these two specific citations by Checkstyle occur only on the doc comment of the method following a method that is annotated as `@generated NOT`. Oddly, not on that annotated method. This cannot be a coincidence

This is surprising and unnecessary because in general I would ignore Checkstyle on generated code. What shall I do about this?

tortmayr:
@cdamus AFAIK the `modelserver.edit` package should already exclude generated sources from checkstyle checks. At least at the current master: https://github.com/eclipse-emfcloud/emfcloud-modelserver/blob/7fc1da39770697e93993ff8310ad6ead0c7472e2/bundles/org.eclipse.emfcloud.modelserver.edit/.checkstyle#L14
I guess we should make sure that the maven checkstyle check also excludes the src-gen directories

cdamus:
Thanks @tortmayr . It looks like this `.checkstyle` config is already trying to exclude the src-gen folder:

```xml
 <file-match-pattern match-pattern="src-gen" include-pattern="false"/>
```

I suppose that isn't working? Or it's used only by the Eclipse Checkstyle plug-in?

tortmayr:
> I suppose that isn't working? Or it's used only by the Eclipse Checkstyle plug-in?

It is working but only for the Eclipse checkstyle plugin. The maven plugin needs be configured separately.
e.g:
```xml
	<plugin>
		<groupId>org.apache.maven.plugins</groupId>
		<artifactId>maven-checkstyle-plugin</artifactId>
		<version>${checkstyle.plugin}</version>
		<configuration>
			<configLocation>emfcloud-checkstyle-8.44.xml</configLocation>
			<consoleOutput>true</consoleOutput>
			<excludes>**/src-gen/**/*.java</excludes>
		</configuration>
		...
       <&plugin>
 ```

cdamus:
Well, for whatever reason, no variation of `<excludes>` configuration that I tried actually caused Checkstyle to skip the files in the `src-gen/` tree:

- `<excludes>**/src-gen/**/*.java</excludes>`
- `<excludes>**/src-gen/**/*</excludes>`
- `<excludes>src-gen/**/*</excludes>`
- `<excludes>**/src-gen/**</excludes>`
- `<excludes>src-gen/**</excludes>`
- etc.

So, in commit 471de656bda2f34e976bd0cc4501fb4084a9cdca I resorted to a suppression filter to skip Javadoc checks on these files that it seems Checkstyle just insists on checking.

ndoschek (approved):
Thanks a lot Christian for the update! Looks great to me! :tada:

Issue #201, opened 2022-04-05
modelserver.example build failure NoClassDefFound org/eclipse/osgi/internal/messages/Msg


gdaniel:
Building the latest version of the repository leads to a build failure on `modelserver.example` with the following error:
```
Exception in thread "Timer-2" java.lang.NoClassDefFoundError: org/eclipse/osgi/internal/messages/Msg
```

How to reproduce: clone the repository and run `mvn compile`.

Complete stack trace
```
Exception in thread "main" java.lang.AssertionError
        at jdk.compiler/com.sun.tools.javac.util.Assert.error(Assert.java:155)
        at jdk.compiler/com.sun.tools.javac.util.Assert.check(Assert.java:46)
        at jdk.compiler/com.sun.tools.javac.comp.Modules.enter(Modules.java:247)
        at jdk.compiler/com.sun.tools.javac.main.JavaCompiler.readSourceFile(JavaCompiler.java:837)
        at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment$ImplicitCompleter.complete(JavacProcessingEnvironment.java:1535)
        at jdk.compiler/com.sun.tools.javac.code.Symbol.complete(Symbol.java:642)
        at jdk.compiler/com.sun.tools.javac.code.Symbol$ClassSymbol.complete(Symbol.java:1326)
        at jdk.compiler/com.sun.tools.javac.code.Type$ClassType.complete(Type.java:1140)
        at jdk.compiler/com.sun.tools.javac.code.Type$ClassType.getTypeArguments(Type.java:1066)
        at jdk.compiler/com.sun.tools.javac.code.Printer.visitClassType(Printer.java:237)
        at jdk.compiler/com.sun.tools.javac.code.Printer.visitClassType(Printer.java:52)
        at jdk.compiler/com.sun.tools.javac.code.Type$ClassType.accept(Type.java:993)
        at jdk.compiler/com.sun.tools.javac.code.Printer.visit(Printer.java:136)
        at jdk.compiler/com.sun.tools.javac.util.AbstractDiagnosticFormatter.formatArgument(AbstractDiagnosticFormatter.java:199)
        at jdk.compiler/com.sun.tools.javac.util.AbstractDiagnosticFormatter.formatArguments(AbstractDiagnosticFormatter.java:167)
[...]

PR #202, opened 2022-04-06
#201 modelserver.example build failure NoClassDefFound

ndoschek:
- Add org.eclipse.osgi to taretplatform


.github/workflows/build.yml
```diff
@@ -9,13 +9,19 @@ jobs:
...
 
     runs-on: ${{ matrix.os }}
     timeout-minutes: 20
 
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+      
+      - name: Set up Maven
+        uses: stCarolas/setup-maven@v4.3
```
martin-fleck-at, line 22: This is an action from a market place and I am not sure we should use it just like that. @eneufeld What is our policy regarding non-verified Github actions? Alternatively, we may need to do the setup ourselves in a custom step.

martin-fleck-at (commented):
Great fix, Nina! Changes look good to me and the build seems to run through as well! However, I am not sure we can use the non-verified GitHub action to setup maven.

.github/workflows/build.yml
ndoschek: I just checked what's else out there, maybe we could use the [Apache Maven Wrapper Plugin](https://maven.apache.org/wrapper/maven-wrapper-plugin/) for that as well. I will try that.

martin-fleck-at (approved):
Fantastic, I think the solution with the maven wrapper is very nice, thank you!

Issue #159, opened 2022-02-02
API V2: Properly define and support multi-resources patches/operations


CamilleLetavernier:
Follow-up issue for #150 (Model Server API V2)

Json Patch typically applies on a single object graph (the current document). However, with EMF, we may have multiple resources in the same resource set, and changes may apply to several of them at the same time.

We need to define how we can reference these resources in Patch operations, and how Json Patch are dispatched to listeners. Currently, we used non-standard Json Patch Path attributes in Json Patch Operations: `resource-id#emf-id/featureName`. However, this won't work with automatically-generated Json Patches.

Once the Patch/Command has been applied, we also need to notify all the model listeners, for all affected resources. Currently, we only send notifications for the main resource (as identified by the `?modeluri` parameter).

CamilleLetavernier:
For "pure Json Patches", I think we should remove the Resource URI from the Operation. These Json Patches should only affect one resource, corresponding to the `?modeluri` parameter. So instead of our custom Json Patch:

```
PATCH ...models?modeluri=mymodel
BODY { op: "replace", path:"mymodel#object/name", value: "New Name" }
```

We'd have:

```
PATCH ...models?modeluri=mymodel
BODY { op: "replace", path:"path/to/object/name", value: "New Name" }
```

Which is a standard Json Patch. In general, if the path includes a # symbol, we can treat it as resourceURI#path, and if it doesn't, we can use the `?modeluri` parameter as the Resource URI, and treat the path as a standard Json Path.

For notifications, we have two cases:

- Model Listeners: each listener is installed on a single modelURI, so we just need to make sure we notify all of them (For each affected resource)
- Request reply: once the command is executed, we return a single Json Patch to the client, as a reply to the request. We should return a map of patches, and indicate the affected model URI for each. For example, replace:

```
{
   data: {
      message: "OK",
      patch: [{ op: "replace", path: "path/to/object/feature", value: "New value"}]
   }
}
```

With:

```
{
   data: {
      message: "OK",
      patches: {
          "file:/firstModel.xmi": [{ op: "replace", path: "path/to/object/feature", value: "New value"}],
          "file:/secondModel.model": [{ op: "replace", path: "path/to/other/object/feature", value: "Another value"}]
      }
   }
}
```

(Note: this map-format could also be used during edition, for patching multiple resources in a single request/transaction).

CamilleLetavernier:
[...]

PR #206, opened 2022-04-08
API v2: Subscription option for URI fragments in update notifications

### LINKED ISSUE #205: API v2: Subscription option for URI fragments in update notifications
The new incremental model update notifications in API v2 that provide a description of changes as JSON Patches are great. However, the `path` of each operation is supplied as a JSON Pointer. For many applications, this is great and convenient for non-EMFish processing of models.

In my Theia application I use UUIDs exclusively in my EMF resources because in editing scenarios the default path-based URI fragments can change radically even for objects that are not otherwise affected by an edit operation. And throughout the application, various components keep track of objects by their IDs, which being UUIDs are stable. And all of these components have information only about the specific objects that they track ("selections", such as in a Properties view). They do not have the whole model in which the JSON Pointers in patches can be resolved relative to the model root object to find the objects that were changed.

This makes it difficult in a component of my application that is subscribed to model updates to detect when an update includes changes in the "selection" in that component. It would be very easy if the operations in the patch referenced objects not by JSON Pointer but by the "custom path" that the Model Server supports in patches sent in `/api/v2/models` `PATCH` requests. That "custom path" being the object's URI fragment plus the feature and optional index.

So, this feature request asks for a new subscription option that instructs the server to provide JSON Patches in which the operations' `path` properties are URI-fragment-based paths and not standard JSON Pointers. I propose a new URL query parameter `paths` with two recognized case-insensitive values:

- `jsonPointer` for standard JSON Pointer paths (the default if no `paths` parameter is supplied)
- `uriFragments` for the "custom" URI-fragment-based paths

cdamus:
As described in the feature request, add a new subscription parameter for `paths` scheme.
Now, with the `paths=uriFragments` query parameter in the `/api/v2/subscribe` request, the `Operation`s in JSON Patches provided by the server in incremental update notifications will have `path` properties of the form

`#_0uuid6017-fun/feature/3`

whereas otherwise they have the form

`/path/1/through/3/containment/0/references/7/to/0/a/1/feature/-`

Note that the resource URI is not included; only the fragment portion. A future enhancement can include resources URIs when they are necessary either to disambiguate operations on objects in different resources or when the resource URI is different to the `modeluri` parameter of the subscription. This is for now on par with the JSON Pointer scheme, which also does not support multiple resources (see #159 for more on that).

Also incidentally fix a compiler warning introduced by an earlier PR.

Resolves #205

### DIFF
--- a/README.md
+++ b/README.md
@@ -396,6 +396,9 @@ public interface ModelServerClientApiV2<A> {
 
    void subscribeWithValidation(String modelUri, SubscriptionListener subscriptionListener, String format,
       long timeout);
+      
+   void subscribe(String modelUri, SubscriptionListener subscriptionListener, SubscriptionOptions options);
+
 
    boolean send(String modelUri, String message);
 
@@ -407,6 +410,52 @@ public interface ModelServerClientApiV2<A> {
 }
 ```
 
+<details>
+	<summary>v2 Subscription Options API</summary>
+
+```java
+public interface SubscriptionOptions extends Serializable {
+
+   String getFormat();
+
+   boolean isLiveValidation();
+
+   long getTimeout();
+
+   String getPathScheme();
+
+   Map<String, String> getAdditionalOptions();
+
+   default boolean hasAdditionalOptions() {
+      return !getAdditionalOptions().isEmpty();
+   }
+
+   static Builder builder() {
+      return new Impl.Builder();
+   }
+
+   interface Builder {
+
+      Builder withFormat(String format);
+
+      Builder withLiveValidation();
+
+      Builder withLiveValidation(boolean validation);
+
+      Builder withTimeout(final long timeout);
+
+      Builder withTimeout(long timeout, TimeUnit unit);
+
+      Builder withPathScheme(String pathScheme);
+
+      Builder withOption(String key, String value);
+
+      SubscriptionOptions build();
+   }
+
+}
+```
+</details>
 
 ### REST API Example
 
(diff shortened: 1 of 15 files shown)


bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/util/JsonPatchHelper.java
```diff
@@ -157,4 +159,28 @@ protected void rollback(final Transaction tx, final ModelServerEditingDomain dom
...
+    */
+   public Optional<String> toURIFragmentPath(final String modelURI, final ResourceSet resourceSet, final String path) {
+      try {
+         SettingValue setting = getSetting(modelURI, resourceSet, path);
+         EObject owner = setting.getEObject();
+         String featureName = setting.getFeature().getName();
+         String fragment = owner.eResource().getURIFragment(owner);
+         String featurePath = setting.getIndex().map(index -> String.format("%s/%s", featureName, index))
+            .orElse(featureName);
+         return Optional.of(String.format("#%s/%s", fragment, featurePath));
```
ndoschek, line 179: I'm wondering if we really need the `#` here?

ndoschek (commented):
Thanks @cdamus, the changes look great to me!
I tested the uri fragment subscription with a simple javascript client and it works as expected!

I would have one request/wish though. Could you update the Java client (and possibly the Example client) as well please? That would be great, but if there's no time left I would also be fine with opening a follow-up issue.

cdamus:
Thanks again @ndoschek. As discussed, commit 0fe3af44ff13015bc4d479ffb7b0844b211b3af8 fixes the problem of the `path` in the transformed JSON Patch operations not including the resource part of the URI, which doesn't match the kind of patches that the server accepts.

cdamus:
And commit 68724d88abc49a1319dd12b32c0a97b70f2e8fc8 adds the new subscription option to the V2 client API and also to the example CLI app.

As the number of subscription options is growing, it seems appropriate to add a more generic API to capture them all together in one bundle in one subscription call without further proliferating overloads of the subscribe method signature.

ndoschek (changes requested):
Thanks again @cdamus, the update of the PR is terrific!
I tested again with my custom client and now the updated Example client and this is definitely a major improvement for all subscribers! :smile: :tada: 

I just have one request left, could you please update the README with the new argument and the `SubscriptionOptions`? Thanks a lot!

cdamus:
Good point, @ndoschek . That's now down in commit f2e231f27ed2884c7021a843193afd2fb4f2024e.

README.md
```diff
@@ -407,6 +410,51 @@ public interface ModelServerClientApiV2<A> {
 }
 ```
 
+<details>
+	<summary>v2 Subscription Options API</summary>
+```java
```
ndoschek, line 415:
```suggestion
	<summary>v2 Subscription Options API</summary>

```java
```

cdamus: Weird, it works fine in Marked.app! Which md previewer do you use, so I can avoid this next time?

I force-pushed commit 281c63400c95c562f64187878f9f39560402f85b over the previous commit to fix this. Thanks!

ndoschek (approved):
Thanks again @cdamus!

PR #208, opened 2022-04-11
API V2: The patch operation "remove" doesn't remove cross references

### LINKED ISSUE #207: API V2: The patch operation "remove" doesn't remove cross references
When using the Patch operation "remove" to delete an object from the model, cross references are not unset. This is a bug in the JsonPatchHelper, that returns a "Remove" command instead of a "Delete" command.

CamilleLetavernier:
- Use DeleteCommand instead of RemoveCommand, when deleting an object (Remove from list and Unset feature still rely on RemoveCommand or SetCommand and won't modify cross-references)

fixes #207


cdamus (approved):
Looks good to me!

Issue #209, opened 2022-04-12
API V2: Return patches for each affected resource


CamilleLetavernier:
This is related to Issue #159, although it can be done independently.

At the moment, when executing a Command or a Patch, we return a Json Patch describing the actual changes applied to to the model. However, this Patch only describes changes affecting the main resource (As defined by the `?modeluri` query). So if we execute:

```
PATCH models?modeluri=main.xmi
patch: [{
  "op": "replace",
  "path": "secondary.xmi#objectId/feature"
  ...
}]
```

We will receive an empty patch, because the main resource wasn't modified. Instead of returning a single patch for the main resource, we should return a patch for each affected resource (which may or may not include the main resource). We can use a Json Map instead of a single patch:

```
patches: {
   "main.xmi": [ { "op": ...}],
   "secondary.xmi": [ { "op": ...}]
}
```

or an Array:

```
patches: [{
   modelUri: "main.xmi",
   patch: [ { "op": ...}]
}, {
   modelUri: "secondary.xmi",
   patch: [ { "op": ...}]
}]
```

The scope of this ticket is to define multiple patches in the reply of edit() operations, as well as Transactions. It is not about the ability to specify multiple patches during an edit operation (This will be covered by Issue #159)

Regarding Subscription listeners: each listener is registered for a single resource, so we don't need to change them. However, we need to make sure that listeners of all affected resources are notified; and not just the main resource listeners.

cdamus:
Additionally/alternatively, we may want to consider supporting a `paths=urifragments` query parameter in the `PATCH` request to encode the `Operation::path` property in the reply's patch as EMF-URI-based "custom paths" as per #205 in the incremental update subscriptions.

For EMF-resource-savvy clients this may be enough to distinguish operations on multiple resources? Or, it could work together with a map of patches to provide the object IDs that are convenient for various client-side indexing.

CamilleLetavernier:
> Additionally/alternatively, we may want to consider supporting a `paths=urifragments` query parameter in the `PATCH` request to encode the `Operation::path` property in the reply's patch as EMF-URI-based "custom paths" as per #205 in the incremental update subscriptions.

I like that :+1:

> For EMF-resource-savvy clients this may be enough to distinguish operations on multiple resources?

IDs/URI Fragments are not necessarily unique across resources (They probably are when using UUIDs, and probably aren't when using path-based EMF Fragments or name-based IDs)

[...]

Issue #214, opened 2022-05-04
API V2: Fix `delete object via patch` request


ndoschek:
The integration test `delete with path - object` fails since PR #208 with the following error.

```
  Integration tests for ModelServerClientV2
    test requests
      ✔ delete with patch - index based (64ms)
      1) delete with patch - object


  1 passing (89ms)
  1 failing

  1) Integration tests for ModelServerClientV2
       test requests
         delete with patch - object:
     AssertionError: expected [ { …(4) } ] to be undefined
      at Context.<anonymous> (src/model-server-client-v2-integration.spec.ts:165:54)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)

```

PR #215, opened 2022-05-06
GH-209: API V2: Return patches for each affected resource

CamilleLetavernier:
Change the return value of edit, undo and redo operations, to return one patch per modified resource, in addition to the main-resource patch.

Before:

```json
{
    "type": "success",
    "data": {
        "message": "Model 'SuperBrewer3000.coffee' successfully updated",
        "patch": [
            {
                "op": "remove",
                "path": "/workflows/0/nodes"
            }
        ]
    }
}
```
After:
```json
{
    "type": "success",
    "data": {
        "message": "Model 'SuperBrewer3000.coffee' successfully updated",
        "patch": [
            {
                "op": "remove",
                "path": "/workflows/0/nodes"
            }
        ],
        "allPatches": [
            {
                "modelUri": "file:/home/camille/Git/modelserver/examples/org.eclipse.emfcloud.modelserver.example/.temp/workspace/SuperBrewer3000.coffee",
                "patch": [
                    {
                        "op": "remove",
                        "path": "/workflows/0/nodes"
                    }
                ]
            }
        ]
    }
}
```

Note that for the main resource, the patches are redundant. We could keep "allPatches" exclusively, but it would make everything a bit more complicated: with a main patch, we don't have to worry about modelUri at all in case of single-resource edition. With multiple-resources, we can rely on allPatches exclusively and ignore the main patch. Keeping both also means we don't break existing clients (and single-resource implementations do not need to change), which is a lot easier.

### DIFF
--- a/bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/DefaultModelController.java
+++ b/bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/DefaultModelController.java
@@ -11,6 +11,7 @@
 package org.eclipse.emfcloud.modelserver.emf.common;
 
 import static org.eclipse.emfcloud.modelserver.emf.common.util.ContextResponse.accepted;
+import static org.eclipse.emfcloud.modelserver.emf.common.util.ContextResponse.badRequest;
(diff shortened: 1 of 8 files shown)


ndoschek (approved):
Thanks @CamilleLetavernier, changes look great to me and everything works as expected! :+1: 
I also noticed, that with this revision issue #214 is fixed as well, so I added it to be closed with this PR.
Just one minor nitpick: Could you please update the lower part of this PR's description, thanks!

PR #221, opened 2022-05-25
220: API V2: JsonPatchHelper doesn't properly handle 'add' operations for single-valued references

### LINKED ISSUE #220: API V2: JsonPatchHelper doesn't properly handle 'add' operations for single-valued references
When using an "add" Operation with a single-valued EMF Reference, we should use an EMF "SetCommand" instead of an "AddCommand", otherwise the resulting command will not be executable (EMF AddCommand only works with Lists)

CamilleLetavernier:
- Use distinct commands for single-valued and multi-valued references

Fixes #220


cdamus (approved):
I wonder, would this be a good point to start collecting JUnit test cases for these patch-to-command transformation scenarios, for regression testing? There will be more changes down the road, such as for handling an insertion index in `add` operations.

CamilleLetavernier:
> I wonder, would this be a good point to start collecting JUnit test cases for these patch-to-command transformation scenarios, for regression testing?

Definitely. I opened #222 to track this

PR #223, opened 2022-05-25
Add missing jars in lib project

eneufeld:
Without the jars the bundle cannot be consumed by p2 correctly as a
ClassNotFoundException is thrown.

### DIFF
--- a/bundles/org.eclipse.emfcloud.modelserver.lib/build.properties
+++ b/bundles/org.eclipse.emfcloud.modelserver.lib/build.properties
@@ -1,15 +1,4 @@
 bin.includes = META-INF/,\
                .,\
-               lib/annotations-20.1.0.jar,\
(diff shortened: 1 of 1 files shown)


cdamus (approved):
Ship it!

Although I might suggest that this is an opportunity to replace the individual JAR entries in the file with a single `lib/` entry, which would prevent this happening for new libraries in the future. 😀

eneufeld:
I thought about it, but this way we exclude source jars, don't know why we add them in the first place :shrug:

cdamus:
Then we can add a `bin.excludes` entry to exclude the source jars using a `*-sources.jar` pattern 😀

eneufeld:
That is a great hint, I try this.

cdamus (approved):
🥇

Issue #216, opened 2022-05-10
Json and schema are inconsistent when using an EMap reference


vhemery:
See https://github.com/eclipse-emfcloud/emfcloud/discussions/155 for broader context

Use a metamodel with an EMap reference (a multivalued reference with type pointing to an EStringToStringMapEntry Eclass, with `java.util.Map$Entry` instance type name).
The EMap reference is serialized by [emfjson-jackson](https://github.com/eclipse-emfcloud/emfjson-jackson) as an Object. Which is good. (although serializer has some issues)

On the other hand, the generated json schema for EMap references (by `org.eclipse.emfcloud.modelserver.jsonschema.DefaultJsonSchemaConverter.createJsonSchema(EReference, boolean)`) is not consistent with this definition, as it considers the feature as an array of entries objects.
This difference also has impacts on JsonForms when trying to use the model objects along with the json data.

The json schema generation should produce map objects schema, such as (for a map with EString values) :
```
"properties":{
   "testReference":{
      "type":"object","additionalProperties":{"type":"string"}
   }
}
```
With such a schema, the EStringToStringMapEntry Eclass is most probably no longer used. However, I don't think it would be wise to delete it completely from the schema, in case it is used by another reference which may be monovalued (eventually in an external Ecore).
The same logic should also work with other key and value types, and not only with Strings. Key needs however to be an attribute to map with a json String key.

Note I am willing to contribute on this.

PR #227, opened 2022-06-08
#216 Json and schema are inconsistent when using an EMap reference

vhemery:
Generate schema for Map object.

closes https://github.com/eclipse-emfcloud/emfcloud-modelserver/issues/216


ndoschek (approved):
Thanks @vhemery! Code looks good to me! :tada: 
Special thanks for adding another test case :medal_sports: 
I did not test the EMap case in-depth as I do not have a use case on hand, but the type schema generation in general still works as expected!
Thanks again!

Issue #187, opened 2022-03-10
Unify logging output to use log4j2 in Eclipse product


ndoschek:
If the ModelServer is packaged as eclipse product and run as Eclipse `IApplication` (e.g. in the `coffee-editor`) the logging should be unified to use log4j2 exclusively.

Currently, there are issues with different logging output streams from `slf4j` (e.g. the Javalin base uses slf4j logging for the startup info).
Also, if the product is started via the `equinox` launcher from the client, errors are logged regarding the missing `slf4j-simple` dependency.

For the classic java jar product, we could overcome this by adding the [Apache Log4j SLF4J Binding](https://mvnrepository.com/artifact/org.apache.logging.log4j/log4j-slf4j-impl) that redirects slf4j to log4j.

ndoschek:
As it turned out, using the modelserver as Java jar product, there might be a problem with aggregating multiple log configurations. I reworked the log configuration with #229 for jar products.

For the Eclipse products, the problem is a bit more complex, as it seems that in case of p2 the classloader cannot grasp the responsible LoggerFactory/LoggerBinder correctly.
It is possible to redirect the underlying jetty server logs to log4j2, but the javalin logs are not correctly redirected (which are for example the startup infos or the devLogging).
We will have to investigate this a bit further.

ndoschek:
We do not continue the work on trying to fix it for the use case of running the Model Server as eclipse product (via equinox for example).
Alos, trends go more towards pure maven builds and running backend services as JARs (e.g. for the coffee-editor https://github.com/eclipse-emfcloud/coffee-editor/issues/456).

PR #229, opened 2022-07-05
#187 Fix logging configuration for JAR product

ndoschek:
- Update log configuration
  - If file path is given, reconfigure logging based on this configuration
  - Default case configures console logging to level INFO, no file logging
- Add ExampleServerLauncher.launch configuration that uses the example log configuration (console and file logging at level info)
- Remove log configuration files from classpath to avoid unwanted configuration aggregation
- Fix performance warning that causes log42j logging issue if executed in bundled jar
- Switch to import instead of require plugin bundles for log4j dependencies
- Update lib dependencies
  - Do not exclude source libs to improve debugging experiences in p2 use case
- Update README
  - Update logging section
  - Format via markdownlint and remove markdownlint warnings
  - Update import project section and remove outdated information
- targetplatform project
  - Include targetplatform as module of releng parent project to be able to import all projects as maven projects without exception
  - Remove unused targetplatform profile

Part of #187

Please also have a look at https://github.com/eclipse-emfcloud/emfcloud-modelserver-theia/pull/110

### DIFF
--- a/.mvn/custom-settings.xml
+++ b/.mvn/custom-settings.xml
@@ -4,7 +4,7 @@
             <id>wso2-mirror</id>
             <mirrorOf>wso2</mirrorOf>
             <name>WSO2 mirror to unblock.</name>
-            <url>http://dist.wso2.org/maven2/</url>
+            <url>http://maven.wso2.org/nexus/content/groups/wso2-public/</url>
             <blocked>false</blocked>
         </mirror>
     </mirrors>
--- a/pom.xml
+++ b/pom.xml
@@ -377,7 +377,7 @@
 				<repository>
 					<id>wso2</id>
 					<name>WSO2</name>
-					<url>http://dist.wso2.org/maven2/</url>
+					<url>http://maven.wso2.org/nexus/content/groups/wso2-public/</url>
 				</repository>
 			</repositories>
 


bundles/org.eclipse.emfcloud.modelserver.common/META-INF/MANIFEST.MF
```diff
@@ -20,4 +19,5 @@ Export-Package: org.eclipse.emfcloud.modelserver.common,
  org.eclipse.emfcloud.modelserver.common.utils,
  org.eclipse.emfcloud.modelserver.jsonschema
 Bundle-Vendor: EclipseSource
-Import-Package: com.google.inject.multibindings;version="1.3.0"
+Import-Package: com.google.inject.multibindings;version="1.3.0",
+ org.apache.logging.log4j;version="2.17.1"
```
cdamus, line 23: Should this not be a version range?

cdamus (changes requested):
Wow, that is a thorough clean-up for developer usability. Thanks!

I do have a few quibbles in comments in-line. Also, there's a problem in the formatting in the new README:

CleanShot 2022-07-07 at 18 25 45@2x

This section lost the fence around the code block and all of the indentation within it.

bundles/org.eclipse.emfcloud.modelserver.common/META-INF/MANIFEST.MF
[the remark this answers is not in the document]
ndoschek: That's a good point, thanks for catching that!

ndoschek:
Thanks a lot for you review @cdamus, I addressed your comments with a separate commit, would be great if you could have another quick look, thanks again!

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/launch/ModelServerLauncher.java
```diff
@@ -125,10 +124,10 @@ protected void configure() {
       });
    }
 
-   public static void configureLogger(final Optional<String> configurationFilePath) {
-      if (configurationFilePath.isPresent()) {
+   protected static void configureLogger(final String configurationFilePath) {
+      if (configurationFilePath != null && !configurationFilePath.isEmpty()) {
```
cdamus, line 128: Good catch on the present-but-empty string! 😀

Issue #167, opened 2022-02-09
Allow EMF Tests to run in p2 build


martin-fleck-at:
Currently we do not build the `org.eclipse.emfcloud.modelserver.emf.tests` tests when building for p2 (`mvn clean verify -Pp2`) as there was some issue with Mockito not being able to mock certain classes. Running the tests from Eclipse, however, works fine.

The goal of this task is to ensure that the tests can also be run with the p2 maven build.

PR #231, opened 2022-07-08
Improve Codec handling for multi module use

### LINKED ISSUE #230: Support for multiple modules
I have a use case where I want to be able multiple modelservers standalone as well as run them as one bis model server. 

Currently the modelserver expects to have exactly one module that extends the ModelServerModule.
Also many of the services are singletons. These services cannot be overriden then and need to be unified manually.

I currently see 2 ways of doing it:
- use private modules which only expose the ModelController using the file extension they support as annotation. This might be problematic for glsp integration where we need to handle multiple resources. Unfortunately a private module cannot add to a multibinding directly. There are workarounds but the elements need to be manually retrieved and pushed into the multibinding. 
- instead of registring codecs in a map using the format, use a provider that also takes the modeluri into account. This probably solves most of the requirements but some special cases would need to be unified manually or also extracted to multibindings.

eneufeld:
Currently Codecs are only registered using the format they support.
This is not enough if different codecs are needed for the same
format but different models.
The new API uses a CodecProvider that can be used with a bazaar pattern
so that the best fitting CodecProvider can be retrieved.

Fix #230


bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/codecs/CodecProvider.java
```diff
@@ -0,0 +1,27 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR MIT
+ ********************************************************************************/
+package org.eclipse.emfcloud.modelserver.emf.common.codecs;
+
+import java.util.Optional;
+import java.util.Set;
+
+import org.eclipse.emfcloud.modelserver.common.codecs.Codec;
+
+public interface CodecProvider {
```
cdamus, line 18: This interface needs some Javadoc.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/codecs/CodecProvider.java
```diff
@@ -0,0 +1,27 @@
...
+
+   int NOT_SUPPORTED = -1;
+
+   Set<String> getAllFormats();
```
cdamus, line 22: What is the purpose of this? Javadoc would help.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/codecs/DefaultCodecsProvider.java
```diff
@@ -0,0 +1,49 @@
...
+import com.google.common.collect.Sets;
+
+public class DefaultCodecsProvider implements CodecProvider {
+
+   private final Set<String> supportedFormats = Sets.newHashSet(ModelServerPathParametersV2.FORMAT_XMI,
+      ModelServerPathParametersV2.FORMAT_JSON, ModelServerPathParametersV2.FORMAT_JSON_V2);
+
+   @Override
+   public Optional<Codec> getCodec(final String modelUri, final String format) {
+      if (ModelServerPathParametersV2.FORMAT_XMI.equals(format)) {
```
cdamus, line 29: This could be a `switch` statement. Or the `supportedFormats` set could instead me a map whose values are `Supplier<Code>` initialized via `XmiCode::new` etc. so that this method just has to get the supplier from the map and call it.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/di/DefaultModelServerModule.java
```diff
@@ -250,7 +250,8 @@ protected void configureMultiBindings() {
       configure(MultiBinding.create(EPackageConfiguration.class), this::configureEPackages);
       configure(MultiBinding.create(Routing.class), this::configureRoutings);
       configure(MapBinding.create(EntryPointType.class, AppEntryPoint.class), this::configureAppEntryPoints);
-      configure(MapBinding.create(String.class, Codec.class), this::configureCodecs);
+      // configure(MapBinding.create(String.class, Codec.class), this::configureCodecs);
```
cdamus, line 253: Delete this line.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/patch/EMFCommandHandler.java
```diff
@@ -71,7 +71,7 @@ private <C> PatchCommand<CCommand> getPatchCommand(final C ctx, final JsonNode p
 
    @FunctionalInterface
    private interface Decoder<C> {
-      Optional<EObject> decode(C context, String payload) throws DecodingException;
+      Optional<EObject> decode(String modelUri, C context, String payload) throws DecodingException;
```
cdamus, line 74: I think a doc comment to indicate that the `modelUri` can be `null` would be helpful. It isn't at all obvious that it would be valid to request a decode for a `null` model URI.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/util/JsonPatchHelper.java
```diff
@@ -80,13 +83,19 @@ public class JsonPatchHelper extends AbstractJsonPatchHelper {
     */
    private static final AtomicBoolean NEED_OBJECT_URI_MAPPINGS = new AtomicBoolean();
 
+   private Optional<CodecProvider> getBestCodecProvider(final String modelUri, final String format) {
+      return this.codecProvider.stream()
+         .sorted((cp1, cp2) -> cp2.getPriority(modelUri, format) - cp1.getPriority(modelUri, format))
+         .findFirst();
```
cdamus, line 89:
```suggestion
         .min((cp1, cp2) -> cp2.getPriority(modelUri, format) - cp1.getPriority(modelUri, format));
```

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/util/JsonPatchHelper.java
```diff
@@ -80,13 +83,19 @@ public class JsonPatchHelper extends AbstractJsonPatchHelper {
     */
    private static final AtomicBoolean NEED_OBJECT_URI_MAPPINGS = new AtomicBoolean();
 
+   private Optional<CodecProvider> getBestCodecProvider(final String modelUri, final String format) {
```
cdamus, line 86: Could this instead be a static method of the `CodecProvider` interface? Something like `CodecProvider::getBestCodecProvider(Collection<? extends CodecProvider>, String, String): Optional<CodecProvider>`.

And, in fact, since usually the next step is to get the codec, it could be nice also to have `CodecProvider::getBestCodec(Collection<? extends CodecProvider>, String, String): Optional<Codec>`. I would expect other clients than the `JsonPatchHelper` could benefit from this.

cdamus (changes requested):
Thanks @eneufeld for the contribution! This is a big improvement over the current design.

cdamus:
Oops, sorry, I didn't notice that this PR was just a draft and now I cannot dismiss my review because it's a draft.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/codecs/DICodecsManager.java
```diff
@@ -66,52 +60,60 @@ public DICodecsManager(final Map<String, Codec> codecs) {
...
+   public Map<String, JsonNode> encode(final String modelUri, final EObject eObject) throws EncodingException {
       Map<String, JsonNode> encodings = new LinkedHashMap<>();
-      formatToCodec.forEach((key, codec) -> {
-         try {
-            encodings.put(key, codec.encode(eObject));
-         } catch (EncodingException e) {
-            e.printStackTrace();
-         }
+      this.codecProviders.forEach(cp -> {
+         cp.getAllFormats().forEach(f -> {
```
cdamus, line 66: Sorry, I missed this in my previous review. We're not considering format priorities, here. We're iterating all available formats from all known providers, but we don't know that we're getting the most appropriate implementation of each format because we can have multiple providers offering the same format.

I think we need to first collect the unique formats supported by all providers, then for each, hold an auction with all providers to get the best codec for the format.

eneufeld:
good catches, I will add those, thank you!

ndoschek:
Thanks Eugen! I didn't have time to look into it yet, but regarding the build, the EMF tests are failing that are only included in the `m2` build (due to known issues in the `p2` environment, see also #167).

eneufeld:
Good thing we have the tests running ;-) . 
Thank you for seeing this.
Updated the test mock setup to fix the test.

cdamus (approved):
Seems to work. Thanks!

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/patch/EMFCommandHandler.java
ndoschek: Perhaps make it even more obvious like that?
```suggestion
      Optional<EObject> decode(@Nullable String modelUri, C context, String payload) throws DecodingException;
```

ndoschek (changes requested):
Thanks a lot Eugen! Overall it looks good to me, I added some minor nitpicks inline.
Would be great if you could have a look, thanks!

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/patch/EMFCommandHandler.java
cdamus: That could be cool, yes! I wasn't sure whether we are using these annotations in this project. I think at least that they are not consistently applied.

ndoschek: Yes I agree, until now they are rarely used, we should probably pursue this further from now on :)

eneufeld: I checked this quickly, we use this in 4 classes in 10 places, so definitely not that often.
Also which annotation to use, there is no standard `javax.annotation` for this. In the ten cases we use the jetbrain one, but there is also an eclipse one `org.eclipse.jdt.annotation.NonNull` which can be use in eclipse null analysis which we did not enable.
I would suggest to create a ticket to discuss this and enable null analysis and unify the annotations. Do you agree?

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/codecs/CodecsManager.java
```diff
@@ -44,41 +44,45 @@ public interface CodecsManager {
...
+   JsonNode encode(String modelUri, WsContext context, EObject eObject) throws EncodingException;
 
    /**
     * Decode a JsonNode to an EObject.
     *
-    * @param context the javalin http context
-    * @param payload tthe String payload holding EObject definition to decode
+    * @param modelUri The ModelUri of the request, null if not available
+    * @param context  the javalin http context
+    * @param payload  tthe String payload holding EObject definition to decode
```
ndoschek, line 81:
```suggestion
    * @param payload  the String payload holding EObject definition to decode
```
and the other occurrences in this file.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/watchers/ModelWatcher.java
```diff
@@ -22,7 +22,7 @@ public interface ModelWatcher {
    /**
     * Factory to create a {@link ModelWatcher} from a supported {@link Resource}.
     */
-   public interface Factory {
+   interface Factory {
```
ndoschek, line 25: Why is this change necessary?

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/codecs/CodecProvider.java
```diff
@@ -0,0 +1,64 @@
...
+   /**
+    * The codec this provider can return for the provided modelUri and format.
+    *
+    * @param modelUri The modelUri to get the codec for
+    * @param format   The format to get the codec for
+    * @return the codec this provider offers for the modelUri, format combination, an empty optional if not supported
+    */
+   Optional<Codec> getCodec(String modelUri, String format);
+
+   static Optional<? extends CodecProvider> getBestCodecProvider(
```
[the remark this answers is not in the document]
ndoschek: Fine by me

ndoschek (commented):
Thanks for the update Eugen! Looks good to me, two final minors from my side. Thanks!

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/watchers/ModelWatcher.java
eneufeld: public is unnecessary for an inter interface of a public interface. JDT complained ;-).

ndoschek (approved):
Thanks a lot Eugen! :+1:

PR #243, opened 2022-08-03
242: EMFCommandHandler requests codec for null model URI

### LINKED ISSUE #242: EMFCommandHandler requests codec for null model URI
This bug is a follow-up to issue #230 and PR #231 in which the `CodecProvider` framework was introduced to support model-specific format mappings.

The problem is that the `EMFCommandHandler` requests a codec for the `null` model URI, but it has the request/web-socket context from which it can get the model URI, so it should use that to get the most appropriate codec available.

cdamus:
Use the model URI provided by the context to request the most appropriate codec.

Fixes #242

### DIFF
--- a/tests/org.eclipse.emfcloud.modelserver.emf.tests/src/org/eclipse/emfcloud/modelserver/emf/patch/EMFCommandHandlerTest.java
+++ b/tests/org.eclipse.emfcloud.modelserver.emf.tests/src/org/eclipse/emfcloud/modelserver/emf/patch/EMFCommandHandlerTest.java
@@ -1,0 +1,102 @@
+/********************************************************************************
+ * Copyright (c) 2022 STMicroelectronics and others.
+ *
+ * This program and the accompanying materials are made available under the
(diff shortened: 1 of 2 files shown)


tests/org.eclipse.emfcloud.modelserver.emf.tests/src/org/eclipse/emfcloud/modelserver/emf/patch/EMFCommandHandlerTest.java
```diff
@@ -0,0 +1,102 @@
+/********************************************************************************
+ * Copyright (c) 2022 EclipseSource and others.
```
ndoschek, line 2:
```suggestion
 * Copyright (c) 2022 STMicroelectronics and others.
```

ndoschek (approved):
Thanks @cdamus!
Looks good to me, special thanks for adding test cases! :sparkles: 
I just have one minor remark about the header (please see inline). Thanks again for the quick fix!

PR #244, opened 2022-08-17
Do not reuse object mapper

sgraband:
Reusing the object mapper can lead to problems.
For example the validation dynamically adds the resource to allow for id mapping.
Hence, a new object mapper is needed each time as the resource might change.


martin-fleck-at (approved):
Looks good to me, thank you Simon!

PR #246, opened 2022-08-22
245: Codecs looked up by workspace URI

### LINKED ISSUE #245: Codecs looked up by workspace URI
This problem is related to the new `CodecProvider` API introduced with #231.

In my application I am finding that, in some flows, the decoding of a JSON Document tries to look up a codec using the workspace URI, which locates a directory, instead of the contextual model URI. Because my codec provider matches on resource URIs (not the workspace location, which may contain heterogeneous resource types), the correct codec is not used.

The problem is that the `DICodecsManager` implementations of the `decode(...)` method variants that accept a workspace URI for resolution of proxy URIs relative to the workspace location use this workspace URI to look up the codec to apply. This despite having also the request or socket context in which to find the model URI.

The workspace URI is certain to be the wrong URI to use as the whole point of the `CodecProvider` API is to support multiple codecs for the same `format` request/session parameter in a workspace, distinguished by model URI.

cdamus:
Use the model URI provided by the request or session context if it is available.

Incidentally fix problems found in the `DefaultModelResourceManager` while debugging the issue in my application:

- catching all throwables, including `Error`s
- NPE in attempt to load a non-existent resource because the resource set is `null`

Fixes #245

### DIFF
--- a/bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/codecs/DICodecsManager.java
+++ b/bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/codecs/DICodecsManager.java
@@ -27,6 +27,7 @@ import org.eclipse.emfcloud.modelserver.common.ModelServerPathParametersV1;
 import org.eclipse.emfcloud.modelserver.common.codecs.Codec;
 import org.eclipse.emfcloud.modelserver.common.codecs.DecodingException;
 import org.eclipse.emfcloud.modelserver.common.codecs.EncodingException;
+import org.eclipse.emfcloud.modelserver.emf.common.ModelURIConverter;
 import org.eclipse.emfcloud.modelserver.emf.common.util.ContextRequest;
 
 import com.fasterxml.jackson.databind.JsonNode;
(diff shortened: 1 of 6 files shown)


bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/codecs/DICodecsManager.java
```diff
@@ -90,7 +93,7 @@ public JsonNode encode(final String modelUri, final Context context, final EObje
    @Override
    public JsonNode encode(final String modelUri, final WsContext context, final EObject eObject)
       throws EncodingException {
-      return findCodec(modelUri, context.queryParamMap()).encode(eObject);
+      return findCodec(modelUri, context).encode(eObject);
```
ndoschek, line 96: Why does it seem that we handle the format argument differently for `encode`/`decode`?
As far as I can see, the two `findCodec` methods are basically doing the exactly same thing, but one is a little more nested than the other. We always expect a format parameter, and if this is not given we return either the `preferredFormat` or `ModelServerPathParametersV1.FORMAT_JSON` which result in the same String.

Can we optimize that and only go with the public method `findCodec(final String modelUri, final WsContext context)` or do you see any problems with that?

ndoschek (commented):
Thanks a lot @cdamus!
Code looks good to me, special thanks for adding another set of tests :sparkles: 

I added one minor comment inline about the retrieval of the format parameter, would be great if you could have a look and tell me what you think about that. Thanks!

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/codecs/DICodecsManager.java
cdamus: That makes sense, yes, thanks. I already made a similar change elsewhere to align the finding of the codec in the WS session case.

(I don't know why there isn't a parallel API for the REST request context case, but that's out of scope, I'm afraid)

cdamus:
Hi @ndoschek commit 8eafdcc0a96661b6e76aec84fde4da6bdbb638d0 makes an attempt to rationalize the looking up of formats from the context query parameters between the request context and socket context cases. With compatibility for existing of overrides of now-deprecated API. What do you think? TIA.

ndoschek (approved):
Thanks @cdamus, looks great to me! Thanks again for the additional effort!

PR #251, opened 2022-09-16
Concurrency conflicts between Transaction- and ModelController

### LINKED ISSUE #250: Concurrency conflicts between Transaction- and ModelController
The TransactionController can interact with the same resource sets, esp. for JSON (de)serialization, as the ModelController, on any number of different threads for incoming requests.

Whereas the ModelController by default has a wrapper that serializes all access onto a single thread, we need the same for the TransactionController. Moreover the ModelController and the TransactionController need to be coördinated on _the same thread_.

As the number of asynchronous requests coming in from client applications via the [Model Server (Node)](/eclipse-emfcloud/modelserver-node) increases, the likelihood of ConcurrentModificationExceptions in the JSON (de)serialization increases with it, as well as problems reading an inconsistent state of a model being requested on `GET /api/v2/models` while some edit transaction is making changes to it.

cdamus:
Extract the single-thread execution mechanism from the `SingleThreadModelController` and generalize it, rebuilding it on the Java `ExecutorService`, as an injectable `ModelSynchronizer` service that is now also used by a new `SingleThreadTransactionController`.

Bind the new `SingleThreadTransactionController` by default.

Fixes #250

### DIFF
--- a/bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/DefaultModelSynchronizer.java
+++ b/bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/DefaultModelSynchronizer.java
@@ -1,0 +1,127 @@
+/********************************************************************************
+ * Copyright (c) 2022 STMicroelectronics and others.
+ *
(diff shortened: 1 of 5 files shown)


CamilleLetavernier (approved):
Looks good to me, it works great!

PR #252, opened 2022-09-20
Fix deadlock on re-entrant sync calls

### LINKED ISSUE #250: Concurrency conflicts between Transaction- and ModelController

cdamus:
Cover the case in which the client of the ModelSynchronizer makes re-entrant syncExec or syncCall invocations.

Fixes #250

### DIFF
--- a/bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/DefaultModelSynchronizer.java
+++ b/bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/DefaultModelSynchronizer.java
@@ -31,10 +31,12 @@ public final class DefaultModelSynchronizer implements ModelSynchronizer {
    private final String name = DefaultModelSynchronizer.class.getSimpleName() + "-" + COUNTER.incrementAndGet();
    private final Logger log = LogManager.getLogger(name);
 
+   private volatile Thread execThread;
+
    private final ExecutorService executor = Executors.newSingleThreadExecutor(runnable -> {
-      Thread thread = new Thread(runnable, name);
-      thread.setDaemon(true);
-      return thread;
+      execThread = new Thread(runnable, name);
+      execThread.setDaemon(true);
+      return execThread;
    });
 
    private final Semaphore executionPermits = new Semaphore(100);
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/DefaultModelSynchronizer.java
```diff
@@ -31,10 +31,12 @@ public final class DefaultModelSynchronizer implements ModelSynchronizer {
    private final String name = DefaultModelSynchronizer.class.getSimpleName() + "-" + COUNTER.incrementAndGet();
    private final Logger log = LogManager.getLogger(name);
 
+   private volatile Thread execThread;
```
cdamus, line 34: This could use a full name.

bundles/org.eclipse.emfcloud.modelserver.emf/src/org/eclipse/emfcloud/modelserver/emf/common/DefaultModelSynchronizer.java
```diff
@@ -43,9 +45,18 @@ public DefaultModelSynchronizer() {
...
    }
 
+   public void dispose() {
+      executor.shutdown();
+   }
+
    @Override
    public void syncExec(final Runnable action) {
-      postAndWait(callable(action));
+      if (Thread.currentThread() == execThread) {
```
cdamus, line 54: For clarity it would be nice to factor this out into a dedicated method.

CamilleLetavernier (approved):
Thanks for the quick fix, looks good to me :+1:

PR #264, opened 2023-01-11
#167 Allow EMF Tests to run in p2 build

### LINKED ISSUE #167: Allow EMF Tests to run in p2 build

ndoschek:
- Update to mockito 4.9.0
  - Due to the update of the mockito versions, we are now able to run the emf tests again also in the p2 build
  - New version is not available via orbit, consume missing via org.eclipse.emfcloud.modelserver.tests bundle
  - Add mockito dependency libs to tests bundle
- If emf.tests are running via p2 the test resources are saved differently, hence let's execute tests on resources in a .temp directory to avoid having git changes on each p2 build.

Resolves #167


eneufeld (approved):
Thank you

tests/org.eclipse.emfcloud.modelserver.emf.tests/src/org/eclipse/emfcloud/modelserver/emf/AbstractResourceTest.java
```diff
@@ -21,12 +26,23 @@
...
    @SuppressWarnings({ "checkstyle:VisibilityModifier" })
    protected ResourceSetImpl resourceSet; // needed in ResourceManagerTest.java
 
+   @BeforeClass
+   public static void setupTestResources() throws IOException {
+      // copy test resources to a temporary resource location to avoid
+      // git changes if executing tests due to saving resources
+      File sourceDirectory = new File(TEST_RESOURCES_PATH);
+      File destinationDirectory = new File(RESOURCE_PATH);
+      copyDirectoryFiles(sourceDirectory, destinationDirectory);
```
eneufeld, line 43: on a second thought why not use the `Files` api of nio ?
see here: https://www.baeldung.com/java-copy-file#nio2

ndoschek: Thanks for the hint, I used now `FileUtils.copyDirectory` from the apache commons lib to cleanup the custom code.

eneufeld (approved):

eneufeld (approved):

PR #265, opened 2023-01-11
#203 Build breaks with Maven version 3.8.5

### LINKED ISSUE #203: Build breaks with Maven version `3.8.5`
With the latest Maven verison `3.8.5` the `p2` build fails with a `NoClassDefFound` (org/eclipse/osgi/internal/messages/Msg) Exception (see also #201).

Currently, we [set the maven version for GitHub actions](https://github.com/eclipse-emfcloud/emfcloud-modelserver/blob/master/.github/workflows/build.yml) explicitly to `3.8.4`, and for the Jenkins build the maven version is currently also `3.8.4`.

We need to investigate, if this problem affects only the `3.8.5` and if not, how we can overcome this problem.

Please also update the README after this was resolved.

ndoschek:
- Ensure build works with maven versions except 3.8.5
- Remove maven wrapper
- Use https url for eclipse release downloads and wso2 repository
- Update README

Resolves #203

Remark: Update deploy and release jobs after approval (ensure maven version and remove maven wrapper)


Jenkinsfile
```diff
@@ -2,7 +2,7 @@ pipeline {
     agent any
 
     tools {
-        maven 'apache-maven-latest'
+        maven 'apache-maven-3.8.6'
```
eneufeld, line 5: is latest not >3.8.6 ?

eneufeld (approved):
Thank you

Jenkinsfile
ndoschek: No, not yet, it is still 3.8.5 - see also https://wiki.eclipse.org/Jenkins#Apache_Maven

eneufeld (approved):

PR #268, opened 2023-02-20
Update example build config to produce a cleaner fat jar

eneufeld:


examples/org.eclipse.emfcloud.modelserver.example/pom.xml
```diff
@@ -64,12 +64,12 @@
          <version>${project.version}</version>
          <scope>provided</scope>
       </dependency>
-      <dependency>
+      <!--<dependency>
```
ndoschek, line 67: Why do we remove this lib dependency?

ndoschek (changes requested):
Thanks Eugen, warnings were reduced significantly during the `m2` build.
Also the size of the standalone JAR was decreased by ~4MB (from 26.4MB to 22.1MB).

However, the standalone JAR cannot be started anymore:
```
java -jar org.eclipse.emfcloud.modelserver.example-0.7.0-SNAPSHOT-standalone.jar 
Error: Unable to initialize main class org.eclipse.emfcloud.modelserver.example.ExampleServerLauncher
Caused by: java.lang.NoClassDefFoundError: org/eclipse/emfcloud/modelserver/emf/di/ModelServerModule
```
Could you have another look there, thanks!



<details>
<summary>Regarding the GitHub workflow builds: --> Extracted this to a separate issue: #269 </summary>

It seems that they updated to maven `3.9.0`, which has a breaking change regarding the `maven.config` file we use: https://maven.apache.org/docs/3.9.0/release-notes.html#potentially-breaking-core-changes
To fix this for the newest maven version, we need to split the arguments in separate lines:

`.mvn/maven.config:`
```
--settings
./.mvn/custom-settings.xml
```
I quickly checked with an older version (`3.8.6`), the adjusted `maven.config` worked for me locally as well.
</details>

examples/org.eclipse.emfcloud.modelserver.example/pom.xml
eneufeld: it's not needed as it is pulled in by the other dependencies. We get rid of many warnings by removing it.

ndoschek:
> Regarding the GitHub workflow builds: --> Extracted this to a separate issue: #269

Workflow are functional again, after rebasing the checks should work again.

